PR #11, opened 2021-09-06
SstOptions from Sstable.rs move to format.proto

ZhongliGao:


engula/src/format/format.proto
```diff
@@ -6,3 +6,4 @@ message FileDesc {
   uint64 file_number = 1;
   uint64 file_size = 2;
 }
+message SstOptions { uint64 block_size = 1;}
```
huachaohuang, line 9: Please format this properly.

huachaohuang (changes requested):
I think the example is broken with this commit. Can you help to confirm and fix it?

huachaohuang (approved):
LGTM

PR #12, opened 2021-09-06
Add BloomFilter

### LINKED ISSUE #8: Implement a BloomFilter
A `Filter` trait has been added without an implementation.
We can add a `BloomFilter` like LevelDB.

Reference: https://github.com/google/leveldb/blob/master/util/bloom.cc

ZhongliGao:
close #8

### DIFF
--- a/engula/src/format/bloom_filter.rs
+++ b/engula/src/format/bloom_filter.rs
@@ -12,6 +12,7 @@ use std::hash::{BuildHasher, Hash, Hasher};
 //  See analysis in [Kirsch,Mitzenmacher 2006].
 //  https://www.eecs.harvard.edu/~michaelm/postscripts/rsa2008.pdf
 pub struct BloomFilter {
+    // bit array
     bitmap: BitVec,
     //size of the bit array
     m: u64,
--- a/engula/src/format/bloom_filter.rs
+++ b/engula/src/format/bloom_filter.rs
@@ -1,0 +1,110 @@
+extern crate bit_vec;
+use crate::format::{FilterBuilder, FilterReader};
+use bit_vec::BitVec;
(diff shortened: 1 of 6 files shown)


ZhongliGao:
close #8

huachaohuang (commented):
Would you please add some comments to explain the algorithm? I don't really understand it at all.

engula/src/format/bloom_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+impl BloomFilter {
+    #[allow(dead_code)]
+    // count : number of elements to insert into bloom filter
+    // false_positive : Bloom filter false_positive rate
+    pub fn new(count: usize, false_positive: f64) -> BloomFilter {
+        let m = m_size(count, false_positive);
+        let k = k_size(false_positive);
+        let hashers = [
```
huachaohuang, line 32: Is that OK to use the same hash function here?

engula/src/format/bloom_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let hash_1 = hasher_1.finish();
+        let hash_2 = hasher_2.finish();
+
+        (hash_1, hash_2)
+    }
+
+    // Get the index from hash value of `k_i`.
+    fn get_index(&self, h1: u64, h2: u64, i: u64) -> usize {
+        (h1.wrapping_add((i).wrapping_mul(h2)) % self.m) as usize
```
huachaohuang, line 80: I think LevelDB's implementation is more efficient. Would you like to optimize it?

huachaohuang (changes requested):
Nice try :)

huachaohuang (approved):
OK, I can merge it now, but I may refactor it later.

PR #21, opened 2021-09-18
add outputs for example

tisonkun:
Signed-off-by: tison <wander4096@gmail.com>

This closes #20 .

PR #22, opened 2021-09-19
fix cargo fmt warnings

tisonkun:

Follow up to #21 .

cc @huachaohuang 

Sorry for this follow up. If we take warnings seriously, we can [require the workflow](https://docs.gearset.com/en/articles/2437757-managing-status-check-rules-in-github) before merging.


huachaohuang (approved):
LGTM

PR #24, opened 2021-09-21
Remove branch filter on workflow trigger strategy

### LINKED ISSUE #23: Remove branch filter on workflow trigger strategy
@huachaohuang @ZhongliGao I don't know why you guys change the branch name, but if it's the case and we have only interest branches on the upstream repository (i.e., no personal branch), we can trigger the workflow on every "push" and "pull request" event.

See also https://docs.github.com/en/actions/reference/events-that-trigger-workflows#configuring-workflow-events .

tisonkun:
This closes #23 .

### DIFF
--- a/.github/workflows/checkcode.yml
+++ b/.github/workflows/checkcode.yml
@@ -1,12 +1,6 @@
 name: Check Code
 
-on:
-  push:
-    branches:
-      - main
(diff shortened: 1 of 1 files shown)


huachaohuang (approved):
LGTM

PR #25, opened 2021-10-05
Reduce binary size

PsiACE:
for x86_64-unknown-linux-gnu

- reduce ~90% size for release
- reduce ~65% size for debug


PsiACE:
Normally, the size on macos is already very small and no additional work is required

tisonkun:
Interesting. @PsiACE could you elaborate a bit with the underneath. I'm curious about how this works.

PsiACE:
I will explain a little later

PsiACE:
Sorry I'm late.

Rust generates an absolutely massive amount of debug info by default. One easy workaround is to compress the debug info sections, on Linux with a `.cargo/config` like the PR. See https://github.com/MaterializeInc/materialize/pull/2691/

> `--compress-debug-sections=zlib-gabi` # instructs the linker to compress debug info, most of the redundant debug info can be well compressed

> `debug = 1` # ask Rust to generate less debug info for release builds, this is usually the right tradeoff for a release build.

Some other optimizations from a tutorial [min-sized-rust](https://github.com/johnthagen/min-sized-rust) . Please check for more information. It is worth mentioning that

> codegen-units = 1 # allow for maximum size reduction optimizations, but it may increase the compilation time

After investigation, it hardly adds much to the compilation time. If necessary, the number of codegen-units can be set for the specific crate, like

> arrow = { codegen-units = 4 }

PsiACE:
> Maybe we can consider it when we have trouble with the binary size.

If it goes well, you'll need it soon. Whether you intend to provide the binary to users or use docker, 90% reduction in release size means only tens of megabytes of space are needed

PsiACE:
> we're starting the normal development cycle and it seems this PR should be rebased onto master.

It conflicts with the main branch, if necessary, I will reopen a new PR

huachaohuang:
@PsiACE thanks for your contribution and explanation, I suggest that we can evaluate this optimization when we have a large enough binary.

PR #62, opened 2021-11-03
docs: add a microunit design draft

huachaohuang:
Try to formalize the design of microunit.


docs/microunit.md
```diff
@@ -0,0 +1,32 @@
...
+This document describes the design of microunit.
+
+Microunit is a lightweight unit orchestration framework for Engula.
+
+## Architecture
+
+![Architecture](images/microunit-architecture.drawio.svg)
+
+A microunit deployment is called a universe.
+A universe consists of multiple nodes, each of which manages a set of units.
```
PsiACE, line 14: Aha. Sounds good. Perhaps the nodes will be called stars, planets, satellites, depending on their role?

docs/microunit.md
```diff
@@ -0,0 +1,32 @@
...
+
+A node is a process running on a virtual or physical machine.
+A node owns a certain amount of resources on the machine.
```
PsiACE, line 17: Will the node have all the resources?

docs/microunit.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Control Plane
+
+A universe is bootstrapped when the first node of the universe starts.
+The first node provisions a control unit to take control of the universe.
+Later, other nodes can communicate with the control unit to join the universe.
+
+For failover, the control unit replicates itself to multiple nodes.
+All control units in a universe form the control plane of the universe.
+The control plane runs a consensus algorithm to elect a leader and replicate metadata.
```
PsiACE, line 29: It looks like this is the control consensus. So is there a similar data consensus for other nodes to each other?

PsiACE: In other words, how to make reliable copies of data nodes in this module?

docs/microunit.md
```diff
@@ -0,0 +1,32 @@
...
+
+The control plane is responsible for managing the universe.
+The control plane pulls information from all nodes to make global decisions about the universe.
```
PsiACE, line 32: I like this idea.

docs/microunit.md
huachaohuang: Well, maybe not, seems we only need nodes and units :)

docs/microunit.md
huachaohuang: It can. But it depends on users. For example, users may choose to allocate only half of the resources on a machine to a node.

docs/microunit.md
PsiACE: Got it. This is my mistake.

PsiACE (approved):
lgtm

docs/microunit.md
```diff
@@ -0,0 +1,32 @@
...
+
+A node is a process running on a virtual or physical machine.
+A node owns a certain amount of resources on the machine.
+A unit is a child process provisioned from a node.
+A unit gets a portion amount of resources from the node.
+
+## Control Plane
+
+A universe is bootstrapped when the first node of the universe starts.
+The first node provisions a control unit to take control of the universe.
```
tisonkun, line 24: Here comes two concrete questions:

1. How does the node know itself the first node? Shall we pass a flag via CLI with "you're the first node" and "you can find the control unit at `$url`"?
2. What if there were multiple control units form a quorum? It seems we should introduce "seed" instances alike concept.

Ref:

* https://etcd.io/docs/v3.4/op-guide/clustering
* https://doc.akka.io/docs/akka/current/typed/cluster.html#joining-configured-seed-nodes

tisonkun: OK I see you talk about it below. You can take this comment and reference inputs for implementation. It won't be a blocker for a draft design.

docs/microunit.md
```diff
@@ -0,0 +1,32 @@
...
+Later, other nodes can communicate with the control unit to join the universe.
+
+For failover, the control unit replicates itself to multiple nodes.
```
tisonkun, line 27: I think in v0.2, we run only one control unit in the universe with a preconfigured discovery strategy. That is, when the second node online, it gets the address of control unit preconfigured, and report itself (with resource it owns) to the control unit.

docs/microunit.md
zojw, line 24: what's the difference between the current version "control unit" and the previous version design's "supreme unit"? (maybe I miss some discussion 😄

zojw: In multi-tenant env, maybe it's useful to management units like control units for each tenant, and one management unit like the supreme unit to coordinate between different tennats~

tisonkun: > what's the difference between the current version "control unit" and the previous version design's "supreme unit"?

The control unit of nodes likes a resource manager of the universe - accept registry of nodes and schedule resources. The supreme unit of a universe is the facade to create/alter/delete databases. The control unit isn't aware concepts of engula (the storage engine) which the supreme unit is.

But yes, they have several overlap description among designs and in the design of demo-1 there is also a "control unit" for engula. We should make a more clear.

docs/microunit.md
huachaohuang: Yes, that's what we are going to deliver in v0.2. But this document is for long-term design.

docs/microunit.md
huachaohuang: Aha, let's forget the unit names in the design of demo-1. I am trying to refactor the design here (including naming). Altought I think the overall idea in demo-1 is still valid.

huachaohuang:
Let's merge this and add more details later.

Issue #59, opened 2021-11-02
Settle down the project layout


huachaohuang:
Rust crates:

- `engula`: the integral library and binary
  - `engula-engine`: contains different kinds of storage engines
  - `engula-kernel`: contains the kernel abstraction and some built-in implementations
  - `engula-journal`: contains the journal abstraction and some built-in implementations
  - `engula-storage`: contains the storage abstraction and some build-in implementations

TODO:

- [ ] Improve descriptions in `Cargo.toml`
  - [x] engula
  - [ ] engula-engine
  - [x] engula-kernel
  - [x] engula-journal
  - [x] engula-storage

huachaohuang:
We have some discussions about the project layout in https://github.com/engula/engula/pull/64#issuecomment-960431302 and https://github.com/engula/engula/pull/64#issuecomment-960433236.

Let me elaborate on the alternative here. We can keep only one top-level crate `engula` and publish it to crates.io. `engula` can contain a lot of internal crates and expose some of them for users. Internal crates are not published so their names can be arbitrary. So the project layout looks like this:

```
engula
   src/
    lib.rs
    api/
      map/
    kernel/
      storage/
    platform/
      object-storage
    microunit/
    ...
```

tisonkun:
@huachaohuang generally I like this layout's looking. For the modularity topic, I'm still investigating "features" mechanism. It seems a conditional compilation approach. I'm unsure whether conditional compilation can handle multiple "modules".

huachaohuang:
Did some experiments about sub-crates. Turns out that it is not going to work if we only publish the top-level crate without the internal crates.

Let's say we have a crate `A` with an internal crate `B`. We can publish only crate `A` without crate `B`. But users of crate `A` will not be able to compile since crate `B` is missing.

So we have two choices: merge everything into one crate or publish all crates at once. I currently prefer the latter. If so, we need to update internal crates, give them a unique name (for example, "journal" will be "engula-journal"), etc.

huachaohuang:
Check [deno dependencies](https://crates.io/crates/deno/1.16.3/dependencies) for an example.

tisonkun:
I'd like to clarify the approaches we can choose:

1. "merge everything into one crate" means like tokio, we publish everything under crate "engula", and users can choose features with feature flags.
2. "publish all crates at once" means like deno, we publish every crate separately, and users can choose crates with their names.

[...]

PR #64, opened 2021-11-03
framework/microunit: add some basic structs

huachaohuang:
Wow, we have a runnable binary with this PR :)

```
USAGE:
    engula node start <ADDR>
```

### DIFF
--- a/src/engula-framework/Cargo.toml
+++ b/src/engula-framework/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "engula-framework"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+microunit = { path = "microunit" }
--- a/src/engula-framework/microunit/src/node_server.rs
+++ b/src/engula-framework/microunit/src/node_server.rs
@@ -12,31 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use clap::{crate_version, Parser};
+use std::net::SocketAddr;
 
-mod node;
+use crate::node::Node;
 
-#[derive(Parser)]
-#[clap(version = crate_version!())]
-struct Command {
-    #[clap(subcommand)]
-    subcmd: SubCommand,
-}
+/// An HTTP server that serves a node.
+pub struct NodeServer {}
 
-impl Command {
-    fn run(&self) {
-        match &self.subcmd {
-            SubCommand::Node(cmd) => cmd.run(),
-        }
+impl NodeServer {
+    pub fn bind(_addr: &SocketAddr) -> NodeServer {
+        NodeServer {}
     }
-}
 
-#[derive(Parser)]
-enum SubCommand {
-    Node(node::Command),
-}
-
-fn main() {
-    let cmd: Command = Command::parse();
-    cmd.run();
+    pub fn serve(&self, _node: Node) {}
 }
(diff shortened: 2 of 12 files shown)


src/engula-framework/microunit/src/node_server.rs
```diff
@@ -0,0 +1,28 @@
...
+
+/// An HTTP server that serves a node.
+pub struct NodeServer {}
+
+impl NodeServer {
+    pub fn bind(_addr: &SocketAddr) -> NodeServer {
+        NodeServer {}
+    }
+
+    pub fn serve(&self, _node: Node) {}
```
tisonkun, line 27: Aha. Dummy implementation.

src/engula-framework/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "engula-framework"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+microunit = { path = "microunit" }
```
PsiACE, line 7: It seems that two modules should be avoided to depend on each other

tisonkun (changes requested):
Can we try to adopt a two level ident package structure and let microunit stay at `src/microunit`?

From https://github.com/engula/engula/pull/62#discussion_r742488203 I found it's hard to distinguish engula (the storage engine) and microunit (which should be unaware of the storage engine logic) if microunit stays at `src/engula-framework/microunit` which leads to an imagine microunit being part of engula.

Also `engula-` prefix seems redundant. You can take [materialize](https://github.com/MaterializeInc/materialize/tree/main/src) as an example.

huachaohuang:
> Can we try to adopt a two level ident package structure and let microunit stay at `src/microunit`?
> 
> From [#62 (comment)](https://github.com/engula/engula/pull/62#discussion_r742488203) I found it's hard to distinguish engula (the storage engine) and microunit (which should be unaware of the storage engine logic) if microunit stays at `src/engula-framework/microunit` which leads to an imagine microunit being part of engula.

So from the perspective of Engula, microunit is part of it, no matter whether microunit is in another crate or repository. However, inside the microunit, it is a standalone crate that can be published alone or moved to another repository.

> 
> Also `engula-` prefix seems redundant. You can take [materialize](https://github.com/MaterializeInc/materialize/tree/main/src) as an example.

A big difference between Engula and other projects like Materialize is that Engula is more a library than a binary. Materialize doesn't need to care about the name of internal crates because those crates are not going to be published. But for Engula, we may want to publish the top-level crates (`engula-xxx` crates) and keep second-level crates internal and unpublished.

huachaohuang:
> A big difference between Engula and other projects like Materialize is that Engula is more a library than a binary. Materialize doesn't need to care about the name of internal crates because those crates are not going to be published. But for Engula, we may want to publish the top-level crates (`engula-xxx` crates) and keep second-level crates internal and unpublished.

Another alternative is to keep only one top-level crate `engula` and make all other crates internal. This alternative is simpler although users have to compile all features of Engula together. Yet another trick we can do is to use conditional compilation with [feature flags](https://doc.rust-lang.org/cargo/reference/features.html).

tisonkun:
@huachaohuang thanks for your explanation. Let me think once more about publishing.

huachaohuang:
Let's move the project layout discussion to https://github.com/engula/engula/issues/59. I think the alternative is worth a try if you think it's easier to understand.

tisonkun (approved):
LGTM. Postpone layout discussion to #59 .

PR #71, opened 2021-11-04
ci: add security audit

PsiACE:
Add basic security audit ci. Powered by cargo-audit.

### DIFF
--- a/.github/workflows/audit-security.yml
+++ b/.github/workflows/audit-security.yml
@@ -1,0 +1,48 @@
+# Copyright 2021 The Engula Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Audit Security
+
+on:
+  push:
+    paths:
+      - '**/**/Cargo.toml'
+  pull_request:
+    paths:
+      - '**/**/Cargo.toml'
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  audit-security:
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, 'audit skip')"
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install cargo-audit
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: cargo-audit
+
+      - name: Generate lockfile
+        uses: actions-rs/cargo@v1
+        with:
+          command: generate-lockfile
+
+      - name: Audit dependencies
+        uses: actions-rs/cargo@v1
+        with:
+          command: audit
--- a/.github/workflows/audit-security.yml
+++ b/.github/workflows/audit-security.yml
@@ -17,17 +17,17 @@ name: Audit Security
 on:
   push:
     paths:
-      - '**/**/Cargo.toml'
+      - '**/Cargo.lock'
   pull_request:
     paths:
-      - '**/**/Cargo.toml'
+      - '**/Cargo.lock'
   schedule:
-    - cron: '0 0 * * *'
+    - cron: '25 4 * * *'
 
 jobs:
   audit-security:
     runs-on: ubuntu-latest
-    if: "!contains(github.event.head_commit.message, 'audit skip')"
+    if: "!contains(github.event.head_commit.message, 'skip audit')"
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/audit-security.yml
+++ b/.github/workflows/audit-security.yml
@@ -37,11 +37,6 @@ jobs:
           command: install
           args: cargo-audit
 
-      - name: Generate lockfile
-        uses: actions-rs/cargo@v1
-        with:
-          command: generate-lockfile
-
       - name: Audit dependencies
         uses: actions-rs/cargo@v1
         with:


PsiACE:
see: https://github.com/PsiACE/engula/runs/4107326355?check_suite_focus=true

tisonkun:
@PsiACE can you provide a link on what `cargo-audit` is for and why we need such a ci task?

tisonkun:
Also I'd prefer start an issue first. The workflow seems not quite necessary from my perspective. We don't have to do something because others do.

PsiACE:
> @PsiACE can you provide a link on what `cargo-audit` is for and why we need such a ci task?

cargo-audit will audit Cargo.lock files for dependencies with security vulnerabilities. see https://rustsec.org/

> Also I'd prefer start an issue first. 

Okay, I agree.

> The workflow seems not quite necessary from my perspective. We don't have to do something because others do. 

Just to make it easier to avoid security issues by removing unmaintained crates and adopting patched versions. I think it will help maintainers save effort and improve security. 

> It's always overhead on GitHub Actions runners.

This overhead is acceptable, and I limit how often it can be run.

tisonkun:
@PsiACE thanks for your explanation! I'll take a look at the details. The next time you may create an issue for it.

.github/workflows/audit-security.yml
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Audit Security
+
+on:
+  push:
+    paths:
+      - '**/**/Cargo.toml'
```
tisonkun, line 20:
```suggestion
      - '**/Cargo.lock'
```

If we audit lock file, watch exactly on lock files. Also if we generate lock file at root only, you may replace with `Cargo.lock`.

.github/workflows/audit-security.yml
```diff
@@ -0,0 +1,48 @@
...
+  pull_request:
+    paths:
+      - '**/**/Cargo.toml'
```
tisonkun, line 23:
```suggestion
      - '**/Cargo.lock'
```

ditto

.github/workflows/audit-security.yml
```diff
@@ -0,0 +1,48 @@
...
+  schedule:
+    - cron: '0 0 * * *'
```
tisonkun, line 25:
```suggestion
    - cron: '25 4 * * *'
```

I assume many crontab job set as `'0 0 * * *'`. As we use public runner, I tend to use a random time.

Also maybe it's enough to run once a week. But I don't have a strong feeling on frequency.

.github/workflows/audit-security.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  audit-security:
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, 'audit skip')"
```
tisonkun, line 30:
```suggestion
    if: "!contains(github.event.head_commit.message, 'skip audit')"
```

Is there an idiom with `'audit skip'`? GitHub uses `'skip ci'` to skip CI and I'd like to follow the pattern.

.github/workflows/audit-security.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Install cargo-audit
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: cargo-audit
+
+      - name: Generate lockfile
+        uses: actions-rs/cargo@v1
+        with:
+          command: generate-lockfile
```
tisonkun, line 43: We already checkin an updated lock file. Can we skip this step?

https://github.com/engula/engula/blob/37c47ac2d15329fedbbd35e5b960b8987a39550f/.github/workflows/checkcode.yml#L79-L100

tisonkun (commented):
Thanks for you contribution! Comments inline.

.github/workflows/audit-security.yml
tisonkun: OK I get your idea that check on toml and generate lock file below. It's another approach and more self-contained.

Either way `**/**/` can be replaced in `**/`

PsiACE:
Sorry, I'm not at the computer, I'll commit during the day tomorrow

.github/workflows/audit-security.yml
PsiACE: ok, you are right, lock file is enough

.github/workflows/audit-security.yml
PsiACE: once a week is enough

PsiACE: > As we use public runner, I tend to use a random time.

Good idea, from experience

.github/workflows/audit-security.yml
PsiACE: ok

.github/workflows/audit-security.yml
PsiACE: I think it's OK, but it needs to be checked

.github/workflows/audit-security.yml
```diff
@@ -0,0 +1,48 @@
...
+  pull_request:
+    paths:
+      - '**/Cargo.lock'
+  schedule:
+    - cron: '25 4 * * *'
+
+jobs:
+  audit-security:
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, 'skip audit')"
```
huachaohuang, line 30: May I ask what does this means?

.github/workflows/audit-security.yml
huachaohuang: So should we remove this step?

.github/workflows/audit-security.yml
tisonkun: It's a custom logic that if the head commit of a push has a message with `'skip audit'`, the action will be skipped (not run). You may take https://github.com/pingcap/tidb-dev-guide/commit/b035e9919e6e52dad6f3a04566be3fcff70960bf as a `'skip ci'` example.

However, I think this PR itself may be overkilled. I found an action to use out-of-the-box [audit-check](https://github.com/actions-rs/audit-check). We can configure a schedule job run in every week to overcome that GitHub security advisories doesn't cover Rust projects yet and that's enough.

What do you think @PsiACE ?

PsiACE: audit-check is great, it can even open an issue to alert you. but it looks unmaintained, and it is not suitable to run at PR.

Either way is fine, and you can even find both at databend.

.github/workflows/audit-security.yml
tisonkun: @huachaohuang yes as long as we check in the lock file.

@PsiACE I see that audit-check is unmaintained, would you mind to reduce this step? Then we can move this PR to merge IMO.

huachaohuang: Hmm, if the action relies on this lock file, then I think it is no harm to generate it here?

PsiACE: Done. And I've tested that it works

huachaohuang: I mean, this action can be reused in other places without relying on others to provide the lock file.

huachaohuang: > Done. And I've tested that it works

Aha, fine then.

PsiACE: > I mean, this action can be reused in other places without relying on others to provide the lock file.

But I think the current version is good enough

You can check https://github.com/engula/engula/pull/71/commits/13c311d88cae51c3d7e173830de6c460d29cb235 to get a basic version  .

tisonkun (approved):
LGTM. Thanks for your contribution @PsiACE !

huachaohuang (approved):
LGTM, thanks!

PR #72, opened 2021-11-05
microunit: integrate axum and implement some simple APIs

huachaohuang:
We have an HTTP server that actually does something now :)

## Usage

Start a node:

```
cargo run -p engula -- node start 127.0.0.1:8000
```

Then create a unit with `curl`:

```
curl --header "Content-Type: application/json" --request POST --data '{"kind": "hello"}' http://127.0.0.1:8000/units
```

Then list all units:

```
curl http://127.0.0.1:8000/units
```

Check https://github.com/engula/engula/pull/72/files#diff-668b2c825f579fdebdae0903e9cf9e2a7aa615af9cab7831a592b8b977f5edf9 for the programming API of microunit.

In a typical scenario, the `Unit` implementation will be a client that talks to a server in another process.

### DIFF
--- a/src/engula-framework/microunit/src/node_server.rs
+++ b/src/engula-framework/microunit/src/node_server.rs
@@ -12,17 +12,50 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::net::SocketAddr;
+use std::{net::SocketAddr, sync::Arc};
 
-use crate::node::Node;
+use axum::{
+    extract::Extension, http::StatusCode, response::IntoResponse, routing::get, AddExtensionLayer,
+    Json, Router, Server,
+};
+
+use crate::{
+    node::Node,
+    unit::{UnitDesc, UnitSpec},
+};
 
 /// An HTTP server that serves a node.
-pub struct NodeServer {}
+pub struct NodeServer {
+    addr: SocketAddr,
+}
 
 impl NodeServer {
-    pub fn bind(_addr: &SocketAddr) -> NodeServer {
-        NodeServer {}
+    pub fn bind(addr: SocketAddr) -> NodeServer {
+        NodeServer { addr }
     }
 
-    pub fn serve(&self, _node: Node) {}
+    pub async fn serve(&self, node: Node) {
+        let router = Router::new()
+            .route("/units", get(list_units).post(create_unit))
+            .layer(AddExtensionLayer::new(Arc::new(node)));
+        Server::bind(&self.addr)
+            .serve(router.into_make_service())
+            .await
+            .unwrap();
+    }
+}
+
+async fn list_units(Extension(node): Extension<Arc<Node>>) -> impl IntoResponse {
+    let descs = node.list_units().await;
+    (StatusCode::OK, Json(descs))
+}
+
+async fn create_unit(
+    Json(spec): Json<UnitSpec>,
+    Extension(node): Extension<Arc<Node>>,
+) -> impl IntoResponse {
+    match node.create_unit(spec).await {
+        Ok(desc) => (StatusCode::CREATED, Json(desc)),
+        Err(_) => (StatusCode::INTERNAL_SERVER_ERROR, Json(UnitDesc::default())),
+    }
 }
--- a/src/engula-framework/microunit/src/node_server.rs
+++ b/src/engula-framework/microunit/src/node_server.rs
@@ -18,11 +18,9 @@ use axum::{
     extract::Extension, http::StatusCode, response::IntoResponse, routing::get, AddExtensionLayer,
     Json, Router, Server,
 };
+use serde_json::json;
 
-use crate::{
-    node::Node,
-    unit::{UnitDesc, UnitSpec},
-};
+use crate::{node::Node, unit::UnitSpec};
 
 /// An HTTP server that serves a node.
 pub struct NodeServer {
@@ -55,7 +53,17 @@ async fn create_unit(
     Extension(node): Extension<Arc<Node>>,
 ) -> impl IntoResponse {
     match node.create_unit(spec).await {
-        Ok(desc) => (StatusCode::CREATED, Json(desc)),
-        Err(_) => (StatusCode::INTERNAL_SERVER_ERROR, Json(UnitDesc::default())),
+        Ok(desc) => {
+            let resp = json!({
+                "desc": desc,
+            });
+            (StatusCode::CREATED, Json(resp))
+        }
+        Err(err) => {
+            let resp = json!({
+                "error": err.to_string(),
+            });
+            (StatusCode::INTERNAL_SERVER_ERROR, Json(resp))
+        }
     }
 }
--- a/src/engula/src/hello_unit.rs
+++ b/src/engula/src/hello_unit.rs
@@ -12,33 +12,32 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use clap::{crate_version, Parser};
+use engula_framework::microunit::{async_trait, Result, Unit, UnitBuilder, UnitDesc, UnitSpec};
 
-mod hello_unit;
-mod node;
-
-#[derive(Parser)]
-#[clap(version = crate_version!())]
-struct Command {
-    #[clap(subcommand)]
-    subcmd: SubCommand,
+pub struct HelloUnit {
+    id: String,
 }
 
-impl Command {
-    async fn run(&self) {
-        match &self.subcmd {
-            SubCommand::Node(cmd) => cmd.run().await,
+#[async_trait]
+impl Unit for HelloUnit {
+    async fn desc(&self) -> UnitDesc {
+        UnitDesc {
+            id: self.id.clone(),
         }
     }
 }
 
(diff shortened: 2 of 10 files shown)


tisonkun:
Looks great! I'll take a closer review by the end of this weekend.

huachaohuang:
@tisonkun Sure, seems I need to grant you some permission first.

tisonkun:
@huachaohuang fine. I will record it by myself :P

[assignee](https://docs.github.com/en/issues/tracking-your-work-with-issues/assigning-issues-and-pull-requests-to-other-github-users) can be anyone who comments on the issue/PR but [reviewer](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) must be a member, weird.

PatrickNicholas (approved):

huachaohuang:
@PatrickNicholas Thanks for your review :)

src/engula/src/hello_unit.rs
```diff
@@ -0,0 +1,43 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use engula_framework::microunit::{async_trait, Result, Unit, UnitBuilder, UnitDesc, UnitSpec};
+
+pub struct HelloUnit {
```
tisonkun, line 17: I'd prefer to use a unit in design with dummy implementation than an example unit in production scope.

tisonkun: BTW why this struct stays in `src/engula` crate? What's this crate for? Now it contains both CLI binary and unit implementation.

src/engula-framework/microunit/src/node_server.rs
```diff
@@ -12,17 +12,50 @@
...
-    pub fn serve(&self, _node: Node) {}
+async fn create_unit(
+    Json(spec): Json<UnitSpec>,
+    Extension(node): Extension<Arc<Node>>,
+) -> impl IntoResponse {
+    match node.create_unit(spec).await {
+        Ok(desc) => (StatusCode::CREATED, Json(desc)),
+        Err(_) => (StatusCode::INTERNAL_SERVER_ERROR, Json(UnitDesc::default())),
+    }
 }
```
tisonkun, line 61: add `delete_unit`?

src/engula-framework/microunit/src/node_server.rs
```diff
@@ -12,17 +12,50 @@
...
+}
 
-    pub fn serve(&self, _node: Node) {}
+async fn create_unit(
+    Json(spec): Json<UnitSpec>,
+    Extension(node): Extension<Arc<Node>>,
+) -> impl IntoResponse {
+    match node.create_unit(spec).await {
+        Ok(desc) => (StatusCode::CREATED, Json(desc)),
+        Err(_) => (StatusCode::INTERNAL_SERVER_ERROR, Json(UnitDesc::default())),
```
tisonkun, line 59: It may be better to return an error code and/or error message on error. I learn this we return `UnitDesc::default()` due to signature restriction, but hope there is a way to fix it.

StatusCode plays an unimportant role in real-world HTTP conversation, and express weaker especially when engula has its own error system.

```
curl --header "Content-Type: application/json" --request POST --data '{"kind": ""}' http://127.0.0.1:8000/units
```

gives `{"id":""}` is confusing.

tisonkun (commented):
A nice demo impl. Comments inline.

src/engula/src/hello_unit.rs
huachaohuang: Consider it a temporary move. It will be removed once we have other units to use here.

src/engula-framework/microunit/src/node_server.rs
huachaohuang: Maybe later, just try to make things work together now.

src/engula-framework/microunit/src/node_server.rs
huachaohuang: Let me try to wrap the unit description in another struct that includes an optional error message.

src/engula-framework/microunit/src/node_server.rs
```diff
@@ -55,7 +53,17 @@ async fn create_unit(
...
+                "desc": desc,
+            });
+            (StatusCode::CREATED, Json(resp))
+        }
+        Err(err) => {
+            let resp = json!({
+                "error": err.to_string(),
+            });
+            (StatusCode::INTERNAL_SERVER_ERROR, Json(resp))
+        }
```
tisonkun, line 67: Nice pattern. We may later review the response schema in detail, but a sum type of either successful response or error message (+error code) looks great.

src/engula-framework/microunit/src/node_server.rs
```diff
@@ -12,17 +12,58 @@
...
+async fn create_unit(
+    Json(spec): Json<UnitSpec>,
+    Extension(node): Extension<Arc<Node>>,
+) -> impl IntoResponse {
+    match node.create_unit(spec).await {
+        Ok(desc) => {
+            let resp = json!({
+                "desc": desc,
+            });
+            (StatusCode::CREATED, Json(resp))
```
huachaohuang, line 60: Maybe we can simply return `desc` here. Let's formalize the response later.

tisonkun (approved):

src/engula-framework/microunit/src/node_server.rs
```diff
@@ -12,17 +12,58 @@
...
+        }
```
zojw, line 61: +1 for http code + different resp body solution, but maybe we need some helper method to avoid works like check result and render resp in each router method, maybe we can introduce some util method like this?


```
async fn list_units(Extension(node): Extension<Arc<Node>>) -> impl IntoResponse {
    JsonResp::new(node.list_units().await)
}

async fn create_unit(
    Json(spec): Json<UnitSpec>,
    Extension(node): Extension<Arc<Node>>,
) -> impl IntoResponse {
    JsonResp::new_with_code(StatusCode::CREATED, node.create_unit(spec).await)
}


struct JsonResp<T, E>(result::Result<T, E>, Option<StatusCode>);

impl <T, E> JsonResp<T, E> {
    fn new(r: result::Result<T, E>) -> Self {
        JsonResp(r, None)
    }

    fn new_with_code(s: StatusCode, r: result::Result<T, E>) -> Self {
        JsonResp(r, Some(s))
    }
}

impl<T, E> IntoResponse for JsonResp<T, E>
where
    T: Serialize,
    E: ToString,
{
    type Body = BoxBody;
    type BodyError = Error;

    fn into_response(self) -> Response<Self::Body> {
        let r = self.0
        .map_err(|e|{
            let resp = json!({
                "error": e.to_string(),
            });
            (self.1.unwrap_or(StatusCode::INTERNAL_SERVER_ERROR), Json(resp))
        })
        .map(|s| {
            let resp = json!({
                "desc": s,
            });
            (self.1.unwrap_or(StatusCode::OK), Json(resp))
        });
        match r {
            Ok(value) => value.into_response().map(box_body),
            Err(err) => err.into_response().map(box_body),
        }
    }
}
```
the idea is force handle logic method(like node.create_units) always return Result<>, then use a utility or middleware(but I didn't find it) in router method to check and render response as ok or error (ps: we also need support pass in status code like POST method to create sth new)

I'm freshman to rust and axum, maybe there is better way to do this~? above code just take a try :)

zojw:
except "error code", how about we think about API-version, it seems often have two chosen:

1. specify the version in the request path, like https://github.com/tokio-rs/axum/blob/ba4d8a2357702d58a00d8306f56b1e5cc7a963db/examples/versioning/src/main.rs#L28
2. specify the version in media types, like https://docs.github.com/en/rest/overview/media-types

maybe we'd better choose one as beginnig~ 😄

src/engula-framework/microunit/src/node_server.rs
huachaohuang: Sounds reasonable. We can build some skeletons when the code base is settled down.

huachaohuang:
> except "error code", how about we think about API-version, it seems often have two chosen:
> 
> 1. specify the version in the request path, like https://github.com/tokio-rs/axum/blob/ba4d8a2357702d58a00d8306f56b1e5cc7a963db/examples/versioning/src/main.rs#L28
> 2. specify the version in media types, like https://docs.github.com/en/rest/overview/media-types
> 
> maybe we'd better choose one as beginnig~ 😄

We can also check k8s https://kubernetes.io/docs/reference/using-api/#api-versioning and https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md. We can open a discussion about it.

huachaohuang:
@zojw @tisonkun How about we merge this and improve it step by step? I am going to simplify the project layout described in https://github.com/engula/engula/issues/59.

tisonkun:
@huachaohuang agree.

PR #78, opened 2021-11-08
Simplify the project layout

huachaohuang:
See https://github.com/engula/engula/issues/59 for more details.


tisonkun:
Generally looks good. @huachaohuang would you elaborate a bit how we do publish, i.e., how these modules delivered during a release?

huachaohuang:
> Generally looks good. @huachaohuang would you elaborate a bit how we do publish, i.e., how these modules delivered during a release?
> 
> I don't see a `#[cfg(feature = "...")]` config and get confused.

We only publish the top-level crate `engula` now. `engula` re-export things from internal crates to users. We may decide to publish sub-crates in the future and leverage features for conditional compilation as the project grows. Nevertheless, the top-level crate `engula` remains the same for users.

tisonkun:
... and when re-export, we will use `cfg` to control which to export with a feature flag?

huachaohuang:
> ... and when re-export, we will use `cfg` to control which to export with a feature flag?

Yes, I guess. I haven't done the survey yet, so suggestions are welcome.

tisonkun (approved):
I make an example at https://github.com/tisonkun/engula/tree/features-demo and things look good. You can check out the head commit for more message.

LGTM. May we keep the PR open for a while to see if further feedback.

PsiACE (approved):
LGTM, we can consider merging it, there doesn't seem to be anything special that needs attention.

zojw:
https://github.com/engula/engula/pull/78#pullrequestreview-799883472 I also take try..

but do we need to make engula/engula export instead of export root folder(it seems tokio do the same thing)? maybe it's useful to do the integration test?

tisonkun:
@zojw I agree to consider about the binary `engula` (cli?) and the library part. Another question is that when we turn on parts of features, whether the library size narrowed as much as possible.

cc @huachaohuang

tisonkun:
The problem about library size is confirmed to be fine ([ref](https://users.rust-lang.org/t/are-optional-dependencies-always-included-in-the-final-binary/67078/2?u=tison) & verified locally). The issue that we consider executable & library separately is still valid.

huachaohuang:
@tisonkun @zojw Thanks for your explorations and suggestions about the project layout!

PR #80, opened 2021-11-08
docs: add storage design document

huachaohuang:
Formalize the top-level design of Storage.

Architecture

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -10,20 +10,20 @@ Storage can be used as a standalone component or integrated with Engula.
 Storage divides data into buckets.
 A bucket stores a set of data objects.
 Each bucket has a unique identifier called the bucket name.
-Each object has an identifier that is unique within the bucket called the object name.
+Each object has an object name that is unique within a bucket.
 Objects are immutable once created.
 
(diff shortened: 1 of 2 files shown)


huachaohuang:
Since no comments here and [a simple implementation](https://github.com/engula/engula/pull/86) has landed, I am going to merge this now.

docs/storage.md
```diff
@@ -0,0 +1,55 @@
...
+This document describes the top-level design of Storage.
+
+Storage provides an abstraction to store data objects.
+Storage can be used as a standalone component or integrated with Engula.
+
+## Semantics
+
+Storage divides data into buckets.
+A bucket stores a set of data objects.
+Each bucket has a unique identifier called the bucket name.
```
PatrickNicholas, line 12: It seems that the bucket like a namespace ? So there might exists some situation that  an user have to change the bucket name but keep the exists objects. It would be better that separate bucket name and the unique identifier.

huachaohuang: Do you mean renaming a bucket? Sounds reasonable to me, not sure if most cloud object storages support this feature, thought.

PR #77, opened 2021-11-08
docs: add journal design document

huachaohuang:
Formalize the top-level journal design.

The architecture looks like this:

Architecture

Issue #65, opened 2021-11-03
Roadmap v0.2 - Journal


huachaohuang:
Tasks:

- [x] Design document
- [x] Journal abstraction
- [x] A memory implementation
- [x] #128 
- [x] #117

Little-Wallace:
How we create a journal stream with other params?  If one tablet is ready to split to two tablet, it means that we shall create one new journal for  the new tablet.

huachaohuang:
> How we create a journal stream with other params?

We can allow users to create a stream with some metadata and persist the metadata in `Journal`. Users may also consider storing metadata in `Kernel`. `Kernel` only supports one stream for now, but we will allow users to create multiple streams sooner or later.

PR #82, opened 2021-11-09
journal: add traits and a memory implementation

huachaohuang:
Add some interfaces as described in https://github.com/engula/engula/pull/77.
This PR also implements a memory journal.

Ref: https://github.com/engula/engula/issues/65


zojw (dismissed):
LGTM

src/journal/src/mem/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod mem_journal;
+mod mem_stream;
+
+pub use self::{mem_journal::MemJournal, mem_stream::MemStream};
+
+#[cfg(test)]
+mod tests {
```
tisonkun, line 21: What do you think of testing for engula? We cannot push all tests into one file (as in mod.rs like this), and it's a decision whether or not we mix up production code with testing code in the same file.

tisonkun (dismissed):
LGTM. One question inline.

src/journal/src/mem/mod.rs
huachaohuang: I have no idea yet. Maybe we can learn from other projects.

PatrickNicholas:
LGTM

src/journal/src/stream.rs
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::{async_trait, error::Result, ResultStream};
+
+// TODO: make it generic
```
tisonkun, line 17: What do you mean by "make it generic"? Does it mean:

```rust
pub struct Event<TS> {
    pub ts: TS,
    pub data: Vec<u8>,
}
```

or `Timestamp<T>`? Is there similar stuff in other projects?

cc @huachaohuang

huachaohuang: The `Event<TS>` one, but should be more complicated than that. The rationale is to let users customize the type of the timestamp. Depending on how users implement their databases, the usage of timestamp can be very different. For example, it can be a simple increasing counter, a logical lock, or a hybrid logical lock, which may need different representations.

huachaohuang: But as long as the timestamp is ordered, it should be fine for the journal.

tisonkun: I'll create an issue for it and continue with a further question.

Generally I tend to avoid `TODO` in comment as described in https://github.com/engula/engula/pull/86#discussion_r746499518 . We don't have to set up rules given anyone can always open an issue or discussion to continue the TODO item. However, it's kind of a single of bad smell.

huachaohuang: I agree that an issue is more useful than a `TODO`. But `TODO` is a quick way to mark down something. It is unnatural for me to open an issue talking about something I'm still writing. And the code maybe even deleted in no time. But generally, I think adding issues for TODO items when a PR is going to merge is a good idea.

Issue #68, opened 2021-11-03
Roadmap v0.2 - Storage


huachaohuang:
Tasks:

- [x] Design document
- [x] Storage abstraction
- [x] A mem storage implementation
- [x] A file storage implementation
- [x] A grpc storage implementation

Most tasks have been done, but we still need to refactor some implementations before releasing v0.2.

huachaohuang:
Reference: https://github.com/delta-io/delta-rs/tree/main/rust/src/storage

huachaohuang:
AWS S3 performance guidelines: https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html

huachaohuang:
@zojw Nice! I think you mean https://github.com/engula/engula/pull/86?

zojw:
@huachaohuang emm....., it's my mistake, s/#68/#86/g 😄 

> I'd like to implement a prototype storage traits for aws-s3 & file system after #86 merged.

huachaohuang:
@zojw You may be interested in https://github.com/tokio-rs/tokio-uring. There is no good support for asynchronous random file read so far, as you may need that when implementing the storage.

huachaohuang:
I am going to implement a remote storage service now.

tisonkun:
The ` A remote storage implementation` is finished and can be checked. I suggest @huachaohuang create a subtask issue the next time and let the PR close that subtask issue, which will automatically check the box and that's what GitHub provides.

huachaohuang:
> The ` A remote storage implementation` is finished and can be checked. I suggest @huachaohuang create a subtask issue the next time and let the PR close that subtask issue, which will automatically check the box and that's what GitHub provides.

Actually, only the client part is finished. I think your suggestion makes sense. I will try to apply it. But it may not work well for me now, since the tasks are not so straightforward to be handled step by step so far. For example, I may rewrite a lot of things if necessary.

zojw:
Hi, I'm trying to implement a storage trait for another local storage implement.

it will sit on `src/storage/fs` folder and the initial version should be a simple implementation:

1. user choose a disk path as base folder
2. create bucket as a secondary folder
3. each object saved as an object inside bucket folder

it's simple rough implements, we know real-world object storage has more complex implementations and handles complex detail, but we may not need a full-functional file system, so choose to start from simple and add more functionality as needed.

[...]

PR #86, opened 2021-11-09
storage: add traits and a memory implementation

huachaohuang:
Refactor storage interfaces as described in https://github.com/engula/engula/pull/80.
This PR also provides a memory storage implementation.

More tests may be added later.

Ref: https://github.com/engula/engula/issues/68


zojw (dismissed):
LGTM

zojw:
do we need to present local/remote/external category concepts in code structure?

https://github.com/engula/engula/pull/80/files#diff-1e47e7218df597c79a882aa1c109e42c332234bd8c92e1c493a788f8f64bb328R38-R40

if so do we need to move `mem` to `local/mem` and impl s3 as `external/s3`?

huachaohuang:
> do we need to present local/remote/external category concepts in code structure?
> 
> https://github.com/engula/engula/pull/80/files#diff-1e47e7218df597c79a882aa1c109e42c332234bd8c92e1c493a788f8f64bb328R38-R40
> 
> if so do we need to move `mem` to `local/mem` and impl s3 as `external/s3`?

No, it's not necessary. We can provide some built-in implementations in `storage/`. Users can also provide their implementations outside.

As for cloud platforms, maybe we can consider adding a `platform` crate to put vendor-related things together. For example:

```
src/
  platform/
    aws/
    azure/
```

zojw:
> platform

ok, I'd like to implement s3 in `src/platform/aws/storage` after this merged cc #68

huachaohuang:
@zojw I refactored the file layout a little bit. Specifically, I move storage traits to an `api/` directory to make things more clear. Would you mind reviewing it again?

src/storage/src/api/storage_bucket.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Uploads an object.
+    async fn upload_object(&self, name: &str) -> Box<dyn StorageObjectUploader>;
+
+    /// Deletes an object.
+    async fn delete_object(&self, name: &str) -> Result<()>;
+}
+
+/// An interface to upload an object.
+#[async_trait]
+pub trait StorageObjectUploader {
```
zojw, line 35: maybe this trait is more suitable laid in `storage_object.rs`~~~?

ditto MemObjectUploader

zojw (commented):
reset LGTM

src/storage/src/api/storage_bucket.rs
huachaohuang: But actually `storage_object.rs` doesn't need this, it is the return type of a `StorageBucket` method, so I think it should be considered as a part of `StorageBucket`.

zojw (approved):

huachaohuang:
@zojw Thanks for your review. You may be also interested in https://github.com/engula/engula/pull/82, the implementation is very similar to this.

Let's merge this now.

src/storage/src/api/mod.rs
```diff
@@ -12,7 +12,19 @@
...
+mod error;
+mod storage;
+mod storage_bucket;
+mod storage_object;
 
-#[async_trait]
-pub trait ObjectHandle {}
+pub use async_trait::async_trait;
+
+// TODO: use std::stream::Stream instead
```
tisonkun, line 22: May we create an issue for it and note how to achieve it? TODO in comments can be always an issue in the project which is more actionable and comment-able.

huachaohuang: Make senses, here is the issue https://github.com/engula/engula/issues/91

src/storage/src/mem/mem_bucket.rs
```diff
@@ -0,0 +1,106 @@
...
+        let objects = self.objects.lock().await;
+        let object_names = objects
+            .keys()
+            .cloned()
+            .map(Ok)
+            .collect::<Vec<Result<String>>>();
+        Box::new(stream::iter(object_names))
+    }
+
+    async fn upload_object(&self, name: &str) -> Box<dyn StorageObjectUploader> {
```
zojw, line 60: maybe we need to detect the state of the bucket that owned this object?

e.g. two people A and B
---- A: get bucket "bucket-test" that exists
---- B: empty and delete "bucket-test"
---- A:  try to upload object into "bucket-test"  <=== should return an error, because the object can not download?

huachaohuang: I think it depends. If we mimic file system behaviors on Linux, it is OK to use an opened but deleted directory/file. Once the directory is closed, all data is gone. But I think we should follow the practice of different object storages here.

PR #87, opened 2021-11-09
storage: prototype impl storage traits for aws-s3

zojw:
ref #68 and move almost code in demo-1 to here

it also has some changes for demo-1

- support create/list/delete buckets
- support delete more than one object at the same time
- support simple one part put object

it also proposes to change some trait definitions in #68 

-  maybe separate "object handler" into  "reader" and "write"
   - reader: to handle read logic
   - writer: to handle multipart write logic(write multiple times and finish once)

still, need handle tasks:

- Storage error
- Retry logic
- Select API
- Write two small multi parts report error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,10 +38,136 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
+name = "aws-auth"
+version = "0.1.0"
+source = "git+https://github.com/awslabs/aws-sdk-rust?tag=v0.0.17-alpha#74cd8a14231e1a005a2f554b072fc9782def5519"
+dependencies = [
+ "aws-types",
+ "pin-project",
+ "smithy-async",
+ "smithy-http",
+ "tokio",
+ "tracing",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-endpoint"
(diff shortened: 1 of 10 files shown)


PsiACE:
just for `chrono 0.4.19`, we can add `RUSTSEC-2020-0159` to `.config/audit.toml`, like this:

```toml
[advisories]
ignore = [
    "RUSTSEC-2020-0159",
]
```

tisonkun:
@PsiACE could you elaborate a bit on the reason?

PsiACE:
> @PsiACE could you elaborate a bit on the reason?

chrono is currently an important dependency in rust ecosystem, we cannot avoid passively importing this dependency. But chrono is not as well maintained, we need to wait until it solves the problem before removing this audit ignore.

zojw:
Hi, I'll refine adopt the new API after #86 merged, so close it first.

@PsiACE thanks, I will fix it when this PR reopens.

tisonkun:
@PsiACE I'd like to know why this method works...and notice that the failure give the vulnerability ID of chrono `RUSTSEC-2020-0159` :P

PsiACE:
> @PsiACE I'd like to know why this method works...and notice that the failure give the vulnerability ID of chrono `RUSTSEC-2020-0159` :P

check https://github.com/engula/engula/runs/4153336692?check_suite_focus=true , and get info

Screenshot 2021-11-10 at 10-27-44 storage impl storage traits for aws-s3 · engula engula 5e3a1c9

PR #88, opened 2021-11-09
add cache module

ZhongliGao:


src/cache/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "cache"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+linked-hash-map = "0.5"
```
PsiACE, line 10: The maintenance state of linked-hash-map is not good, you can consider using ritelinked to repalce it.

hashlink is also possible, but I think ritelinked would be better.

ZhongliGao: Thank you for your suggestion. I'll consider replacing it later.

ZhongliGao: > The maintenance state of linked-hash-map is not good, you can consider using ritelinked to repalce it.
> 
> hashlink is also possible, but I think ritelinked would be better.

Thank you for your suggestion. I'll consider replacing it later.

huachaohuang:
@ZhongliGao Thanks for your contribution! You can consider converting this PR to a draft before it is ready. It is also more friendly to provide some description about the PR.

ZhongliGao:
> @ZhongliGao Thanks for your contribution! You can consider converting this PR to a draft before it is ready. It is also more friendly to provide some description about the PR.

ZhongliGao:
> @ZhongliGao Thanks for your contribution! You can consider converting this PR to a draft before it is ready. It is also more friendly to provide some description about the PR.

Ok

PR #93, opened 2021-11-11
storage: refactor file structure and trait names

huachaohuang:
Sorry for moving things in and out, I am still exploring different practices :)

This PR also fixes the `Stream` problem by adding an `Unpin` bound.


huachaohuang:
@zojw You may be interested in this if you are working on storage.

src/storage/src/bucket.rs
```diff
@@ -12,27 +12,27 @@
...
 #[async_trait]
-pub trait StorageBucket {
+pub trait Bucket {
     /// Returns a handle to the object.
-    async fn object(&self, name: &str) -> Result<Box<dyn StorageObject>>;
+    async fn object(&self, name: &str) -> Result<Box<dyn Object>>;
 
     /// Returns a stream of object names.
-    async fn list_objects(&self) -> Stream<Result<String>>;
+    async fn list_objects(&self) -> BoxStream<Result<Vec<String>>>;
```
zojw, line 24: why do we change `BoxStream<Result<String>>` to `BoxStream<Result<Vec<String>>>`? maybe the previous one is more suitable for stream~?

zojw (commented):
reset LGTM

zojw (approved):

PR #94, opened 2021-11-11
docs: add white background for dark theme

huachaohuang:
I found that some images look very bad in dark theme :(


huachaohuang:
This doesn't seem reviewable, let's get it merged to fix those poor images.

PR #98, opened 2021-11-12
storage: add a grpc remote storage

huachaohuang:
This PR only implements the client part of the storage and leaves the server part to future PRs.

This storage uses gRPC for communications. Since the storage APIs are very simple, maybe we can find a more efficient way for RPC in the future. As long as the user interface remains the same, we will be fine.

The API design mostly follows the [Google API design guide](https://cloud.google.com/apis/design) and the [googleapis](https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto), which is very nice and worth reading. However, most part of the current implementation is suboptimal in order to keep things simple for now.

### DIFF
--- a/src/storage/src/grpc/bucket.rs
+++ b/src/storage/src/grpc/bucket.rs
@@ -1,0 +1,109 @@
+// Copyright 2021 The Engula Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use futures::{future, stream};
+
+use super::{
+    client::Client,
+    object::RemoteObject,
+    proto::{DeleteObjectRequest, ListObjectsRequest, UploadObjectRequest},
+};
+use crate::{async_trait, Bucket, Object, ObjectUploader, Result, ResultStream};
+
+pub struct RemoteBucket {
+    client: Client,
+    bucket: String,
+}
+
+impl RemoteBucket {
+    pub fn new(client: Client, bucket: String) -> RemoteBucket {
+        RemoteBucket { client, bucket }
+    }
+}
+
+#[async_trait]
+impl Bucket for RemoteBucket {
+    async fn object(&self, name: &str) -> Result<Box<dyn Object>> {
+        let object = RemoteObject::new(self.client.clone(), self.bucket.clone(), name.to_owned());
+        Ok(Box::new(object))
+    }
+
+    async fn list_objects(&self) -> ResultStream<String> {
+        let input = ListObjectsRequest {
+            bucket: self.bucket.clone(),
+        };
+        match self.client.list_objects(input).await {
+            Ok(output) => {
+                let object_names = output
+                    .objects
+                    .into_iter()
+                    .map(Ok)
+                    .collect::<Vec<Result<String>>>();
+                Box::new(stream::iter(object_names))
+            }
+            Err(err) => Box::new(stream::once(future::err(err.into()))),
+        }
+    }
+
+    async fn upload_object(&self, name: &str) -> Box<dyn ObjectUploader> {
+        let uploader =
+            RemoteObjectUploader::new(self.client.clone(), self.bucket.clone(), name.to_owned());
+        Box::new(uploader)
+    }
+
+    async fn delete_object(&self, name: &str) -> Result<()> {
+        let input = DeleteObjectRequest {
+            bucket: self.bucket.clone(),
+            object: name.to_owned(),
+        };
+        let _ = self.client.delete_object(input).await?;
+        Ok(())
+    }
+}
+
+struct RemoteObjectUploader {
+    client: Client,
+    bucket: String,
+    object: String,
+    content: Vec<u8>,
+}
+
+impl RemoteObjectUploader {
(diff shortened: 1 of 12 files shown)


tisonkun (approved):
The code side looks good to me. I'd like to ask what components will implement as `storage_server`? From the design I can see a "Storage Service" but is it another storage unit?

Also I like `method!` macro and be curious that it isn't implemented in tonic or other libs.

huachaohuang:
@tisonkun The idea is to let different components provide a general type of [tower service](https://docs.rs/tower-service/0.1.0/tower_service/trait.Service.html), which has a good Rust ecosystem. For example, the grpc storage will provide a grpc storage service. And most units will be a general-purpose HTTP server that can serve one or more such kind of services. For example, a storage unit will be an HTTP server that serves a storage service.

So the `Unit/UnitBuilder` trait in microunit will be the bridge between different units and services. Then we may need to decide where these unit implementations should go. For example, we can keep the components totally unaware of microunit by placing unit implementations in a middle layer.

As for the `method!` macro, I think it may only work for now. As the implementation becomes more complicated, different methods may handle things differently. So I think `method!` will be removed sooner or later :)

huachaohuang:
I will try to implement the grpc storage service and unit next to see if these things work together. Then I will update the design documents.

src/storage/src/grpc/bucket.rs
```diff
@@ -0,0 +1,109 @@
...
+        let input = ListObjectsRequest {
+            bucket: self.bucket.clone(),
+        };
+        match self.client.list_objects(input).await {
+            Ok(output) => {
+                let object_names = output
+                    .objects
+                    .into_iter()
+                    .map(Ok)
+                    .collect::<Vec<Result<String>>>();
```
zojw, line 52: we also can remove this line.

src/storage/src/grpc/client.rs
```diff
@@ -0,0 +1,58 @@
...
+impl Client {
+    method!(list_buckets, ListBucketsRequest, ListBucketsResponse);
+
+    method!(create_bucket, CreateBucketRequest, CreateBucketResponse);
+
+    method!(delete_bucket, DeleteBucketRequest, DeleteBucketResponse);
+
+    method!(list_objects, ListObjectsRequest, ListObjectsResponse);
+
+    method!(upload_object, UploadObjectRequest, UploadObjectResponse);
```
zojw, line 46: for upload/read object maybe we can using [grpc stream](https://github.com/hyperium/tonic/blob/master/examples/routeguide-tutorial.md#client-side-streaming-rpc) , but it seems be complex and can do it later~

zojw (approved):
rest LGTM

src/storage/src/grpc/client.rs
huachaohuang: Yes, I want to do that too. But I just try to make things work to see how they fit together for now :)

src/storage/src/grpc/storage.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    async fn list_buckets(&self) -> ResultStream<String> {
+        let input = ListBucketsRequest {};
+        match self.client.list_buckets(input).await {
+            Ok(output) => {
+                let bucket_names = output
+                    .buckets
+                    .into_iter()
+                    .map(Ok)
+                    .collect::<Vec<Result<String>>>();
```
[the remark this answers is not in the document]
huachaohuang: Oh, really!? Let me give it a try.

src/storage/src/grpc/storage.rs
```diff
@@ -0,0 +1,72 @@
...
+            }
+            Err(err) => Box::new(stream::once(future::err(err.into()))),
+        }
+    }
+
+    async fn create_bucket(&self, name: &str) -> Result<Box<dyn Bucket>> {
+        let input = CreateBucketRequest {
+            bucket: name.to_owned(),
+        };
+        let _ = self.client.create_bucket(input).await?;
```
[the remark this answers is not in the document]
huachaohuang: I just assume that it is required to compile. Let me see if I can remove it.

src/storage/src/grpc/object.rs
```diff
@@ -0,0 +1,62 @@
...
+    async fn read_at(&self, buf: &mut [u8], offset: usize) -> Result<usize> {
+        let input = ReadObjectRequest {
+            bucket: self.bucket.clone(),
+            object: self.object.clone(),
+            offset: offset as i64,
+            length: buf.len() as i64,
+        };
+        let output = self.client.read_object(input).await?;
+        let content = output.content;
+        // TODO: refine the interface to avoid copy here.
```
tisonkun, line 57: If the parameter is `&mut [u8]`, we must copy the value from output to the empty buffer.

@huachaohuang do you want to pass a pointer of reference, or is there zero-copy technique to mapping one buffer to another?

huachaohuang: Well, we can change the interface to return a buffer. But in some cases, callers may have a buffer at handed and just want us to copy the data into their buffer, then returning a buffer to them is not friendly. So maybe we need to find a way that satisfies different use cases. Or we simply provide different interfaces. I don't know yet.

zojw: maybe we can 
```
async fn read_at<'a>(&self, offset: usize, len: usize, opt_buf: &'a mut [u8]) -> (Result<usize>, Cow<'a, &[u8]>)
```
and fill and return `opt_buf` if user provide it, and return rpc buffer if not.

but in this way, we cannot metaphor `len` information from buf.len() any more...- -

tisonkun: @huachaohuang OK. You can create an issue post-v0.2. In this case I'd prefer:

1. Provide two interfaces for each use case.
2. Provide one interface returning buffer and let users decide what to do.
3. Provide the current interface with a known and bearable overhead, at least for early releases.

@zojw I don't think an interface in that signature a good idea - it's too complex to understand and anti-intuitive.

PR #103, opened 2021-11-12
journal: make timestamp generic

### LINKED ISSUE #100: journal - make timestamp generic
What do you mean by "make it generic"? Does it mean:

```rust
pub struct Event<TS> {
    pub ts: TS,
    pub data: Vec<u8>,
}
```

or `Timestamp<T>`? Is there similar stuff in other projects?

cc @huachaohuang

_Originally posted by @tisonkun in https://github.com/engula/engula/pull/82#r748146358_

The `Event<TS>` one, but should be more complicated than that. The rationale is to let users customize the type of the timestamp. Depending on how users implement their databases, the usage of timestamp can be very different. For example, it can be a simple increasing counter, a logical lock, or a hybrid logical lock, which may need different representations.

_Originally posted by @huachaohuang in https://github.com/engula/engula/pull/82#discussion_r748150920_

tisonkun:
This closes #100 .

### DIFF
--- a/src/journal/src/journal.rs
+++ b/src/journal/src/journal.rs
@@ -12,21 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::fmt::Display;
-
-use super::{async_trait, error::Result, stream::Stream, ResultStream};
+use super::{async_trait, Result, ResultStream, Stream, Timestamp};
 
 /// An interface to manipulate event streams.
 #[async_trait]
-pub trait Journal<Timestamp: Ord + Display + Send + Clone + Copy> {
+pub trait Journal<TS: Timestamp> {
     /// Returns a handle to the event stream.
-    async fn stream(&self, name: &str) -> Result<Box<dyn Stream<Timestamp>>>;
+    async fn stream(&self, name: &str) -> Result<Box<dyn Stream<TS>>>;
 
     /// Returns a stream of event stream names.
     async fn list_streams(&self) -> ResultStream<String>;
 
     /// Creates an event stream.
-    async fn create_stream(&self, name: &str) -> Result<Box<dyn Stream<Timestamp>>>;
+    async fn create_stream(&self, name: &str) -> Result<Box<dyn Stream<TS>>>;
 
     /// Deletes an event stream.
     async fn delete_stream(&self, name: &str) -> Result<()>;
--- a/src/journal/src/journal.rs
+++ b/src/journal/src/journal.rs
@@ -21,15 +21,15 @@ use super::{
 
 /// An interface to manipulate event streams.
 #[async_trait]
-pub trait Journal<Ts: Timestamp> {
+pub trait Journal<T: Timestamp> {
     /// Returns a handle to the event stream.
-    async fn stream(&self, name: &str) -> Result<Box<dyn Stream<Ts>>>;
+    async fn stream(&self, name: &str) -> Result<Box<dyn Stream<T>>>;
 
     /// Returns a stream of event stream names.
     async fn list_streams(&self) -> ResultStream<String>;
 
     /// Creates an event stream.
-    async fn create_stream(&self, name: &str) -> Result<Box<dyn Stream<Ts>>>;
+    async fn create_stream(&self, name: &str) -> Result<Box<dyn Stream<T>>>;
 
     /// Deletes an event stream.
     async fn delete_stream(&self, name: &str) -> Result<()>;
--- a/src/journal/src/journal.rs
+++ b/src/journal/src/journal.rs
@@ -12,19 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::fmt::Display;
+
 use super::{async_trait, error::Result, stream::Stream, ResultStream};
 
 /// An interface to manipulate event streams.
 #[async_trait]
-pub trait Journal {
+pub trait Journal<Timestamp: Ord + Display + Send + Clone + Copy> {
     /// Returns a handle to the event stream.
-    async fn stream(&self, name: &str) -> Result<Box<dyn Stream>>;
+    async fn stream(&self, name: &str) -> Result<Box<dyn Stream<Timestamp>>>;
 
     /// Returns a stream of event stream names.
     async fn list_streams(&self) -> ResultStream<String>;
 
     /// Creates an event stream.
-    async fn create_stream(&self, name: &str) -> Result<Box<dyn Stream>>;
+    async fn create_stream(&self, name: &str) -> Result<Box<dyn Stream<Timestamp>>>;
 
     /// Deletes an event stream.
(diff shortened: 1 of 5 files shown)


src/journal/src/mem/journal.rs
```diff
@@ -33,8 +33,8 @@ impl Default for MemJournal {
 }
 
 #[async_trait]
-impl Journal for MemJournal {
-    async fn stream(&self, name: &str) -> Result<Box<dyn Stream>> {
+impl<TS: Timestamp + 'static> Journal<TS> for MemJournal<TS> {
```
tisonkun, line 36: I tend to avoid lifetime annotations since they are propagated everywhere quickly. And a `'static` annotation is too restricted from my perspective in this case.

tisonkun: Since we're using `Mutex` already, maybe there is a way to tell the compiler the lifetime bound is satisfied.

tisonkun: From https://discord.com/channels/873004811774214145/873004812684390460/908925007604879360 I'm going to change `Timestamp` with a lifetime bound `+ 'static`.

src/journal/src/journal.rs
```diff
@@ -12,19 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::{async_trait, error::Result, stream::Stream, ResultStream};
+use super::{async_trait, Result, ResultStream, Stream, Timestamp};
```
huachaohuang, line 15: I actually intended to specify the modules here. The rationale is [here](https://github.com/rust-lang/api-guidelines/discussions/255#discussion-3676709). But different practices exist, I am not sure if there are better ways to go.

src/journal/src/journal.rs
```diff
@@ -12,19 +12,19 @@
...
 
 /// An interface to manipulate event streams.
 #[async_trait]
-pub trait Journal {
+pub trait Journal<TS: Timestamp> {
```
huachaohuang, line 19: If we follow the Rust style for acronym, it should be `Ts`. However, I think we can simply use `T`.

src/journal/src/mem/stream.rs
```diff
@@ -20,35 +20,35 @@ use tokio::sync::Mutex;
...
+            .collect::<Vec<Result<Event<TS>>>>();
         Box::new(stream::iter(iter))
     }
 
-    async fn append_event(&self, ts: Timestamp, data: Vec<u8>) -> Result<()> {
+    async fn append_event(&self, ts: TS, data: Vec<u8>) -> Result<()> {
         let event = Event { ts, data };
         let mut events = self.events.lock().await;
-        let last_ts = events.back().map(|x| x.ts).unwrap_or(0);
+        let last_ts = events.back().map(|x| x.ts).unwrap_or_default();
```
huachaohuang, line 51: If we use a `match` or `if let` here, maybe we can eliminate the `Default` trait bound for `TS`.

src/journal/src/stream.rs
```diff
@@ -12,26 +12,29 @@
...
 // limitations under the License.
 
-use super::{async_trait, error::Result, ResultStream};
+use std::fmt::Display;
 
-// TODO: make it generic
-pub type Timestamp = u64;
+use super::{async_trait, Result, ResultStream};
+
+pub trait Timestamp: Ord + Display + Send + Copy + Default + 'static {}
```
huachaohuang, line 19: Maybe we should use `Debug` instead of `Display`. We only print `Timestamp` on errors (do we have to?), so `Debug` should be sufficient. I think `Debug` is more convenient than `Display` because `Display` can not be derived.

tisonkun: Sounds reasonable. I place `Display` here because we use `{}` [here](https://github.com/engula/engula/pull/103/files#diff-8f372267fee5ef26bf79a12dba2f0cea9717a2d41f5b617556f9bb65a668ea8dR54). Update to `Debug`.

src/journal/src/journal.rs
tisonkun: I'd prefer `Ts` which is more descriptive. In my experience `T` always means `Type` where the type parameter of trait is obviously (e.g., List<T>). It's unclear to me what type parameter `Journal` should take.

src/journal/src/journal.rs
tisonkun: OK. Let me revert the changes and learn from the discussion referred.

src/journal/src/mem/stream.rs
tisonkun: I can make it as:

```rust
    async fn append_event(&self, ts: Ts, data: Vec<u8>) -> Result<()> {
        let event = Event { ts, data };
        let mut events = self.events.lock().await;
        if let Some(last_ts) = events.back().map(|x| x.ts) {
            if ts <= last_ts {
                return Err(Error::InvalidArgument(format!(
                    "timestamp {:?} <= last timestamp {:?}",
                    ts, last_ts
                )));
            }
            events.push_back(event);
        }
        Ok(())
    }
```

but it changes semantic. That's what I'm unsure in #100:

> Change `.unwarp_or(0)` to `.unwrap` due to `Ord` itself doesn't provide a reasonable "default" in "last_ts" context. We may introduce a new trait but I don't have a good background on it.

tisonkun:
@huachaohuang comments addressed except https://github.com/engula/engula/pull/103#discussion_r748678388 .

src/journal/src/mem/stream.rs
huachaohuang: I think it's fine. If there is no event yet, any event can be accepted.

src/journal/src/stream.rs
```diff
@@ -12,26 +12,29 @@
...
-// TODO: make it generic
-pub type Timestamp = u64;
+pub trait Timestamp: Ord + Debug + Send + Copy + Default + 'static {}
+
+impl<T> Timestamp for T where T: Ord + Debug + Send + Copy + Default + 'static {}
 
 #[derive(Clone, Debug, Default)]
-pub struct Event {
-    pub ts: Timestamp,
+pub struct Event<Ts: Timestamp> {
```
huachaohuang, line 24: Hmm, I still think a simple `T` is more common and sufficient because we have a trait bound here, after all. But maybe it is personal, so I don't insist here.

src/journal/src/mem/stream.rs
tisonkun: `events.push_back(event)` should be moved outside the `if let` block following your description. Updated.

src/journal/src/stream.rs
tisonkun: Make sense. Given there is one type parameter yet, I'll change `Ts` to `T`.

tisonkun:
@huachaohuang comments addressed. Please give it another look, especially for 033e9ba36db48ff476fd106a41aded67b9b3c7a5.

huachaohuang (approved):
LGTM

PR #107, opened 2021-11-14
journal: refactor traits with associated types

huachaohuang:
The previous implementation used trait objects. But it is less performant and also bring some restriction to implementations (https://github.com/engula/engula/pull/99#discussion_r748134635). Associated types provide more flexibility to users and implementers. For example, implementers can define their error type. On the other hand, users can use more precise types, and if they don't care about those types, they can always erase the types with trait objects.

Note that I also remove the `list_streams` temporarily to simplify things for now, as we may not need it in a while.


src/journal/src/mem/journal.rs
```diff
@@ -33,31 +35,24 @@ impl<T: Timestamp> Default for MemJournal<T> {
...
         let mut streams = self.streams.lock().await;
-        match streams.try_insert(name.to_owned(), stream) {
-            Ok(v) => Ok(Box::new(v.clone())),
-            Err(_) => Err(Error::AlreadyExists(format!("stream '{}'", name))),
+        match streams.entry(name.to_owned()) {
+            hash_map::Entry::Vacant(ent) => {
+                ent.insert(stream.clone());
+                Ok(stream)
+            }
+            hash_map::Entry::Occupied(_) => Err(Error::AlreadyExists(format!("stream '{}'", name))),
```
tisonkun, line 55: Is this a style-only change or has a nontrivial effect?

src/journal/src/stream.rs
```diff
@@ -14,27 +14,32 @@
...
+/// A generic timestamp to order events.
+pub trait Timestamp: Ord + Send + Copy + Debug + Unpin {}
 
-impl<T> Timestamp for T where T: Ord + Debug + Send + Copy + 'static {}
+impl<T: Ord + Send + Copy + Debug + Unpin> Timestamp for T {}
 
-#[derive(Clone, Debug, Default)]
-pub struct Event<T: Timestamp> {
+#[derive(Clone)]
+pub struct Event<T> {
```
tisonkun, line 25: Shall we keep the type bound `T: Timestamp`? Otherwise, it looks like an event with arbitrary type inner data and restriction changed.

tisonkun (commented):
Nice refactor :p

Comments inline and below is a handmade patch to express comments.

```diff
diff --git a/src/journal/src/mem/mod.rs b/src/journal/src/mem/mod.rs
index 942c6a1..950c26c 100644
--- a/src/journal/src/mem/mod.rs
+++ b/src/journal/src/mem/mod.rs
@@ -33,9 +33,15 @@ mod tests {
     async fn test() -> Result<()> {
         let j: MemJournal<u64> = MemJournal::default();
         let stream = j.create_stream("a").await?;
+        stream
+            .append_event(Event {
+                ts: 0,
+                data: vec![0, 1, 2, 3],
+            })
+            .await?;
         let mut events = stream.read_events(0).await;
         while let Some(Ok(event)) = events.next().await {
-            println!("{:?}", event.ts);
+            println!("{:?}", event);
         }
         Ok(())
     }
diff --git a/src/journal/src/mem/stream.rs b/src/journal/src/mem/stream.rs
index 11034a9..3cb43cf 100644
--- a/src/journal/src/mem/stream.rs
+++ b/src/journal/src/mem/stream.rs
@@ -71,12 +71,12 @@ impl<T: Timestamp> Stream for MemStream<T> {
     }
 }
 
-pub struct EventStream<T> {
+pub struct EventStream<T: Timestamp> {
     events: Vec<Event<T>>,
     offset: usize,
 }
 
-impl<T> EventStream<T> {
+impl<T: Timestamp> EventStream<T> {
     fn new(events: Vec<Event<T>>) -> Self {
         EventStream { events, offset: 0 }
     }
diff --git a/src/journal/src/stream.rs b/src/journal/src/stream.rs
index 0d950fa..df013d7 100644
--- a/src/journal/src/stream.rs
+++ b/src/journal/src/stream.rs
@@ -21,8 +21,8 @@ pub trait Timestamp: Ord + Send + Copy + Debug + Unpin {}
 
 impl<T: Ord + Send + Copy + Debug + Unpin> Timestamp for T {}
 
-#[derive(Clone)]
-pub struct Event<T> {
+#[derive(Clone, Debug)]
+pub struct Event<T: Timestamp> {
     pub ts: T,
     pub data: Vec<u8>,
 }
```

src/journal/src/mem/journal.rs
huachaohuang: `try_insert` is an experimental API that needs to be enabled by a feature flag (https://github.com/engula/engula/pull/107/files#diff-50aec1ea2bb95b8166fcb9dd17bc8a0bf606ffcc81f41966af6dd9982dd551b5L15). We can do the same thing in this way to avoid the dependency.

src/journal/src/stream.rs
huachaohuang: I am not sure if it is necessary. But I think when in doubt, choosing a more strict bound first is a good pattern.

tisonkun (approved):
+1 to merge

PR #112, opened 2021-11-17
fix aduit vulnerability and warning

tisonkun:
* fix: the next release version is 0.2.0 (this fixes the warning reported by cargo audit)
* dep: update tokio to fix aduit vulnerability (this fixes the vulnerability reported newly by cargo audit)

@huachaohuang @zojw you can review this PR. It has an overlap with #99 but focuses on fixing all audit reports. It should be OK to rebase one after the other merged first. Or @zojw merge the change on "version = 0.1.0" to "version = 0.2.0" and I can close this PR as integrated in #99 .

This closes #111 .

PR #113, opened 2021-11-17
Add Cahce module

ZhongliGao:
add Cache module as #76  
This PR adds Cache Trait


ZhongliGao:
@huachaohuang you can review this pr. This PR adds Cache Trait

src/cache/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "cache"
+version = "0.1.0"
```
huachaohuang, line 3: Let's bump this to "0.2.0" the same as https://github.com/engula/engula/pull/112.

src/cache/Cargo.toml
```diff
@@ -0,0 +1,9 @@
...
+edition = "2021"
+publish = false
+
+
```
huachaohuang, line 7: This is an unnecessary newline.

src/cache/Cargo.toml
```diff
@@ -0,0 +1,9 @@
...
+[dependencies]
+linked-hash-map = "0.5.4"
```
huachaohuang, line 9: This is unused.

src/cache/src/error.rs
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub struct Error {}
+
+pub type Result<T> = std::result::Result<T, Error>;
```
huachaohuang, line 17: This file is unused.

src/cache/src/cache.rs
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{borrow::Borrow, hash::Hash};
+
+pub trait Cache {
+    type Key: Hash + Eq;
+    type Value;
+
+    fn insert(&mut self, key: Self::Key, value: Self::Value) -> Option<Self::Value>;
```
huachaohuang, line 21: What is the semantic of inserting an existing entry?

src/cache/src/cache.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    fn get<Q: ?Sized>(&mut self, key: &Q) -> Option<&mut Self::Value>
```
huachaohuang, line 23: Why do we return a mutable reference to the value?

huachaohuang (commented):
I think we need to define the semantic of the cache first. For example, should we make the cache thread safe (then it should not take a `&mut self`)?

src/cache/Cargo.toml
ZhongliGao: > This is unused.

The lru implementation will use it

src/cache/src/cache.rs
ZhongliGao: > What is the semantic of inserting an existing entry?

The semantics here is the update of the cache queue, perhaps it would be better to use update for the fn name.

ZhongliGao:
> I think we need to define the semantic of the cache first. For example, should we make the cache thread safe (then it should not take a `&mut self`)?

Update cache, remove and clear caches need &mut self

huachaohuang:
> Update cache, remove and clear caches need &mut self

It's not true. For example, you can use a `Mutex` for this.

src/cache/src/lru_cache.rs
```diff
@@ -0,0 +1,77 @@
...
+        return (true, None);
+    }
+
+    async fn get<Q: ?Sized>(&self, key: &Q) -> Option<Self::Value>
+    where
+        Self::Key: Borrow<Q>,
+        Q: Hash + Eq + Sync + Send,
+    {
+        let map = self.queue.lock().await;
+        map.get(key).cloned()
```
huachaohuang, line 51: I think you need `get_refresh` instead. Otherwise, the map will not be updated.

src/cache/src/cache.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[async_trait]
+pub trait Cache {
+    type Key: Hash + Eq + Sync + Send;
+    type Value: Send + Clone;
+
+    /// Insert an item in the cache.
+    ///
+    /// The first return value indicates whether an insertion has taken place
+    /// (because the cache can refuse to insert an item). The second return
```
huachaohuang, line 27: When do we refuse to insert an item? It is ok for the cache to drop the item, but I think users don't need to care about that. A simpler semantic gives the implementer more chance to optimize.

src/cache/src/lru_cache.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use async_trait::async_trait;
+use linked_hash_map::LinkedHashMap;
+use tokio::sync::Mutex;
+
+use crate::cache::Cache;
+
+pub struct LruCache<K: Hash + Eq + Sync + Send, V: Send + Clone> {
+    queue: Mutex<LinkedHashMap<K, V>>,
+    max_size: usize,
```
huachaohuang, line 27: I think users usually want to control the size bytes of the cache, not the number of elements in the cache.

huachaohuang:
@ZhongliGao It seems there are no activities and responses for this PR. Are you still going to work on it?

ZhongliGao:
> @ZhongliGao It seems there are no activities and responses for this PR. Are you still going to work on it?

I would like to do further optimization of cache and finish it in the near future. I'll start by changing the state to draft.

ZhongliGao:
> @ZhongliGao It seems there are no activities and responses for this PR. Are you still going to work on it?



> @ZhongliGao It seems there are no activities and responses for this PR. Are you still going to work on it?

I want to close this PR, the cache module I did not do very well, many details are poorly implemented. I'm ready to start the design again.

PR #114, opened 2021-11-17
docs: add the design document of Warehouse

huachaohuang:
The semantics of Warehouse is useful to build some analytical databases like Snowflake and DeltaLake.

Warehouse


docs/warehouse.md
```diff
@@ -0,0 +1,28 @@
...
+It is possible that a client fails to upload an object or fails to commit the uploaded objects. In this case, the corresponding objects become obsolete.
+Warehouse relies on garbage collection to purge obsoleted or to be deleted objects in the end.
+
+Warehouse employs a multi-version mechanism to manage metadata.
+It maintains multiple versions of metadata. Each version represents a snapshot of metadata at a specific time.
+Each metadata transaction (add or remove objects) creates a delta version that transforms the last version into a new one.
+When a client connects to Warehouse, it gets the last version from Warehouse as its base version and then subscribes to delta versions.
+When a delta version arrives, the client applies it to the base version to catch up with Warehouse.
+The client can maintain a list of live versions for ongoing queries and release a version once it is no longer required.
+Warehouse guarantees that objects recorded in all client versions remain valid until the corresponding versions are released.
```
zojw, line 28: how do warehouses know "all client candidates"~? maybe the client can be temp leave or be network partition, it seems not simple to know "all live client released the version" :)?

huachaohuang: I think the implementation will be similar to a distributed lock or [leases in etcd](https://etcd.io/docs/v3.3/dev-guide/interacting_v3/#grant-leases). Clients need to communicate with Warehouse to keep their versions alive.

docs/warehouse.md
```diff
@@ -0,0 +1,28 @@
...
+To add objects to Warehouse, a client uploads objects to Storage first and then commits the uploaded objects to Warehouse.
+To delete objects from Warehouse, a client commits the to be deleted objects to Warehouse and relies on Warehouse to delete those objects.
+It is possible that a client fails to upload an object or fails to commit the uploaded objects. In this case, the corresponding objects become obsolete.
+Warehouse relies on garbage collection to purge obsoleted or to be deleted objects in the end.
+
+Warehouse employs a multi-version mechanism to manage metadata.
+It maintains multiple versions of metadata. Each version represents a snapshot of metadata at a specific time.
+Each metadata transaction (add or remove objects) creates a delta version that transforms the last version into a new one.
+When a client connects to Warehouse, it gets the last version from Warehouse as its base version and then subscribes to delta versions.
+When a delta version arrives, the client applies it to the base version to catch up with Warehouse.
```
zojw, line 26: IMO, reverse synced version info just used to let client can release version? how about query & modification arrival client during "un-catch up" time period?  I guess they should not take care about those reverse synced version? (if so maybe read stale data and inconsistent between different warehouse client component)

huachaohuang: I don't think I understand the questions. Specifically, what does "reverse synced version" mean?

docs/warehouse.md
```diff
@@ -0,0 +1,28 @@
...
+- Background jobs that need to add or remove multiple objects atomically to guarantee data correctness
+
+## Architecture
+
+![Architecture](images/warehouse-architecture.drawio.svg)
+
+Warehouse stores object data in Storage and stores object metadata in Manifest.
+
+To add objects to Warehouse, a client uploads objects to Storage first and then commits the uploaded objects to Warehouse.
+To delete objects from Warehouse, a client commits the to be deleted objects to Warehouse and relies on Warehouse to delete those objects.
```
tisonkun, line 18:
```suggestion
To delete objects from Warehouse, a client commits the to-be-deleted objects to Warehouse and relies on Warehouse to delete those objects.
```

tisonkun, line 18: In this way, a client should talk to both Storage and Warehouse. So a user should keep in mind these usages. It seems a bit awkward. I'll imagine with the Warehouse facade, a user can talk to the Warehouse only.

docs/warehouse.md
```diff
@@ -0,0 +1,28 @@
...
+It is possible that a client fails to upload an object or fails to commit the uploaded objects. In this case, the corresponding objects become obsolete.
+Warehouse relies on garbage collection to purge obsoleted or to be deleted objects in the end.
```
tisonkun, line 20:
```suggestion
Warehouse relies on garbage collection to purge obsoleted or to-be-deleted objects eventually.
```

docs/warehouse.md
tisonkun, line 26: So the client will hold a connection to subscribe updates per object or per Warehouse? It seems like something that can help in implementing materialize view. We can think more in this direction.

tisonkun (commented):
Thanks for submitting this PR! Comments inline.

Also `\n` in markdown without a blank line won't start a new line, you can check the rendered version and I think it's the primarily used version that we should take care of. I don't insert a line break without a blank line :P

docs/warehouse.md
```diff
@@ -0,0 +1,28 @@
...
+
+Warehouse is a high-level object storage abstraction based on [Storage](storage.md) and Manifest.
+Warehouse enhances Storage with versioned metadata and atomic metadata operations to serve the following scenarios:
+
+- Foreground queries that need to make sure that their required data remains valid during execution
+- Background jobs that need to add or remove multiple objects atomically to guarantee data correctness
+
+## Architecture
+
+![Architecture](images/warehouse-architecture.drawio.svg)
```
tisonkun, line 13: From the graph, it seems the client talks to Warehouse directly. Will Warehouse becomes a new kind of unit? I'd like to confirm whether it's at the same level of Storage or a component inside Compute.

docs/warehouse.md
huachaohuang: Make senses. We can hide Storage from the abstraction.

docs/warehouse.md
huachaohuang: I think it should be bucket-level. And maybe we should only support bucket-level transactions too.

docs/warehouse.md
huachaohuang: Good question. Warehouse is between Storage and Compute. I am considering running Warehouse and Manifest in the same unit. There are still some designs to be done in the overall architecture and microunit.

docs/warehouse.md
```diff
@@ -0,0 +1,28 @@
...
+
+Warehouse stores object data in Storage and stores object metadata in Manifest.
+
+To add objects to Warehouse, a client uploads objects to Storage first and then commits the uploaded objects to Warehouse.
```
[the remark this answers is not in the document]
huachaohuang: The assumption is that clients will not upload the same object concurrently. This can be achieved by assigning a unique name to each object (for example, a UUID or a sequence number issued by Warehouse).

docs/warehouse.md
huachaohuang: I did consider using database concepts here. But it seems that Warehouse is more like an object management abstraction. So it may be easier to understand if we refer to the bucket/object concepts from Storage.

SGZW: Bucket concepts may cause many clients to fetch many recent metadata updates that it does not need

huachaohuang: Hmm, I think most applications will map one database to one bucket. I need to make it clear that the bucket concept may not actually map to a bucket in the cloud storage. It is ok for Warehouse to map multiple buckets to a single bucket in Storage. Does it sound confused? Maybe I should consider use "database" instead of "bucket" here 🤔

SGZW: Yes, it's will be more clearly, s3's bucket is more familiar with me

huachaohuang:
Thanks for everyone's feedback. While the design is far from perfect, I think it's still better to have it landed and move the design forward. So I am going to merge it now, some wording can be improved later.

docs/warehouse.md
tisonkun: With #140 and #143, a client (engine) is still holding all of:

* Kernel (Warehouse here)
* Stream
* Journal

```rust
#[derive(Clone)]
pub struct Engine<K: Kernel> {
    kernel: K,
    stream: K::Stream,
    bucket: K::Bucket,
    // ...
}
```

PR #115, opened 2021-11-18
storage: implement grpc server

huachaohuang:
This is the server part of https://github.com/engula/engula/pull/98.
The server simply wraps another storage to provide a storage service.
The trait bounds in the server are very cumbersome. But at least it works for now. We can consider refactoring the trait bounds later.

### DIFF
--- a/src/storage/src/grpc/mod.rs
+++ b/src/storage/src/grpc/mod.rs
@@ -31,23 +31,26 @@ pub use self::{
 
 #[cfg(test)]
 mod tests {
+    use tokio::net::TcpListener;
+    use tokio_stream::wrappers::TcpListenerStream;
+
     use super::*;
     use crate::*;
 
-    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[tokio::test(flavor = "multi_thread")]
     async fn test() -> std::result::Result<(), Box<dyn std::error::Error>> {
-        let addr = "127.0.0.1:12345";
+        let listener = TcpListener::bind("127.0.0.1:0").await?;
+        let local_addr = listener.local_addr()?;
         tokio::task::spawn(async move {
-            let addr = addr.parse().unwrap();
             let server = Server::new(mem::MemStorage::default());
             tonic::transport::Server::builder()
                 .add_service(server.into_service())
-                .serve(addr)
+                .serve_with_incoming(TcpListenerStream::new(listener))
                 .await
                 .unwrap();
         });
 
-        let url = format!("http://{}", addr);
+        let url = format!("http://{}", local_addr);
         let storage = RemoteStorage::connect(&url).await?;
         let bucket = storage.create_bucket("bucket").await?;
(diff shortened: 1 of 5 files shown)


tisonkun (approved):
LGTM. Clean implementation! One suggestion on style, not strict but follow the style in `PhantomData` docs.

src/storage/src/grpc/server.rs
```diff
@@ -0,0 +1,123 @@
...
+
+pub struct Server<O, B, S>
+where
+    O: Object,
+    B: Bucket<O>,
+    S: Storage<O, B>,
+{
+    storage: S,
+    _object: PhantomData<O>,
+    _bucket: PhantomData<B>,
```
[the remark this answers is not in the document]
huachaohuang: Can you give me the `PhantomData` link? I don't see it providing any style guide about this. I think prefixing it with an underscore similar to [unused variables](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#unused-variables) makes more sense.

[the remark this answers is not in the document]
tisonkun: I refer to [this example](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#unused-type-parameters). And thanks for clarifying that `_x` means unused vars. From Python conventions, it's more like a "private" field which confuses me in the first place.

src/storage/src/grpc/mod.rs
```diff
@@ -17,11 +17,48 @@ mod client;
...
 };
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::*;
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test() -> std::result::Result<(), Box<dyn std::error::Error>> {
+        let addr = "127.0.0.1:12345";
```
zojw, line 39: https://github.com/hyperium/tonic/issues/667#issuecomment-847847088 it seems rust also can use `:0` test trick 😄

zojw (dismissed):
LGTM

src/storage/src/grpc/mod.rs
huachaohuang: Sounds good. But how do I get the address for the client to connect below 🤔

zojw: we can listen out of spawn block and pass in listener..

```
        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
        let addr = listener.local_addr().unwrap();
        tokio::task::spawn(async move {
            let server = Server::new(mem::MemStorage::default());
            tonic::transport::Server::builder()
                .add_service(server.into_service())
                .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(listener))
                .await
                .unwrap();
        });
```

but it seems dirty that need add new test dependency.. 😞 

```
tokio-stream = {version = "0.1.5", features = ["net"]}
```

PR #118, opened 2021-11-19
layout: separate engula binary and library

### LINKED ISSUE #89: Consider separate executable and library
In https://github.com/engula/engula/pull/78 we simplify the project layout into:

```
├── Cargo.lock
├── Cargo.toml
├── src
│   ├── api
│   ├── background
│   ├── hello_unit.rs
│   ├── main.rs
│   ├── manifest
│   ├── microunit
│   ├── node.rs
│   └── storage
```

However, the root cargo config file define both the executable and library, while a consumer of the library doesn't want the executable part. We may consider releasing executable and library separately. @zojw suggested learn from how tokio does release.

cc @huachaohuang @PsiACE

huachaohuang:
Closes #89 .

We also keep sub-commands in the library so that users can embed them in their binaries easily.


Cargo.toml
```diff
@@ -15,8 +19,7 @@ members = [
...
 
 [dependencies]
-engula-storage = { package = "storage", path = "src/storage" }
-engula-journal = { package = "journal", path = "src/journal" }
-engula-microunit = { package = "microunit", path = "src/microunit" }
+microunit = { path = "src/microunit" }
+
 clap = "3.0.0-beta.5"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
```
tisonkun, line 25:
```suggestion
tokio = { version = "1", features = ["full"] }

```

I'll investigate whether we can add a editorconfig to ensure every file with a newline so that we don't see this kind of blank changes time after time.

tisonkun (dismissed):
LGTM.

Cargo.toml
huachaohuang: Yes, I create an [issue](https://github.com/engula/engula/issues/120) for it.

PR #119, opened 2021-11-19
journal: implement remote journal

### LINKED ISSUE #117: A remote journal implementation
We can wrap a gRPC Journal similar to [the gRPC Storage](https://github.com/engula/engula/tree/main/src/storage/src/grpc). I think it is not very hard, so maybe someone can give it a try :)

tisonkun:
This commit implements for only Timestamp=u64.

This closes #117.


src/journal/src/grpc/stream.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match Pin::new(&mut self.events).poll_next(cx) {
+            Poll::Ready(Some(Ok(resp))) => Poll::Ready(Some(Ok(Event {
+                ts: resp.ts,
+                data: resp.data,
+            }))),
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(None) => Poll::Ready(None),
+            Poll::Ready(Some(Err(s))) => Poll::Ready(Some(Err(Error::from(s)))),
```
tisonkun, line 99: Is it possible to simplify this code? I'm assuming a `map`/`map_err` alike transformation.

cc @huachaohuang @skyzh @leiysky

src/journal/src/grpc/server.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match Pin::new(&mut self.events).poll_next(cx) {
+            Poll::Ready(Some(Ok(event))) => Poll::Ready(Some(Ok(ReadEventResponse {
+                ts: event.ts,
+                data: event.data,
+            }))),
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(None) => Poll::Ready(None),
+            Poll::Ready(Some(Err(err))) => Poll::Ready(Some(Err(Status::from(err)))),
+        }
```
tisonkun, line 152: ditto this snippet.

Also, IntelliJ Rust Plugin complain that this `match` isn't exhausted (false positive) XD

src/journal/src/mem/stream.rs
```diff
@@ -43,10 +43,10 @@ impl<T: Timestamp> Stream for MemStream<T> {
     type EventStream = EventStream<Self::Timestamp>;
     type Timestamp = T;
 
-    async fn read_events(&self, ts: Self::Timestamp) -> Self::EventStream {
+    async fn read_events(&self, ts: Self::Timestamp) -> Result<Self::EventStream> {
```
tisonkun, line 46: read_events remotely can fail

src/journal/Cargo.toml
```diff
@@ -8,4 +8,11 @@ publish = false
 thiserror = "1.0"
 async-trait = "0.1"
 futures = "0.3"
-tokio = { version = "1.13", features = ["full"] }
\ No newline at end of file
+tokio = { version = "1.13", features = ["full"] }
+tokio-stream = { version = "0.1", features = ["net"] }
+tonic = "0.6"
+prost = "0.9"
+serde = { version = "1.0", features = ["derive"] }
```
tisonkun, line 15: I'm thinking of introducing `serde` to enable Timestamp <=> bytes, but `Deserialize<'de>` has a lifetime parameter that I'm unsure how to make the decision. Digging...

src/journal/src/grpc/stream.rs
huachaohuang: I think [`tonic::Streaming`](https://docs.rs/tonic/0.6.1/tonic/struct.Streaming.html#impl-Stream) already implements `futures::Stream`. So maybe you don't need to do this 🤔

tisonkun: The inner type is different:

* `Streaming<ReadEventResponse>`
* `EventStream` with `Item = Result<Event<u64>>`

So I need to do some transformations anyway, by introducing a wrapper, or if there is a transformation method.

huachaohuang: Oh, I see. Maybe take a look at [`StreamExt`](https://docs.rs/futures/0.3.17/futures/stream/trait.StreamExt.html#method.map).

src/journal/src/mem/stream.rs
huachaohuang: You can return a error stream instead. Maybe something like `stream::once(future::err(...))`.

src/journal/src/grpc/stream.rs
tisonkun: Give it a try. Found that `Streaming` has `type Item = Result<T, Status>;` where `Status` is incompatible to `grpc::error:Error`, so we cannot let `EventStream = Streaming<...>`. With `StreamExt` we can make some transformations, but the associated type is more complex.

src/journal/src/mem/stream.rs
tisonkun: `stream::once` gives a `futures::stream::Once` which is incompatible with `journal::grpc::stream::EventStream`.

So we meet a similar issue as `Streaming<...>` vs. `Self::EventStream` which requires adapting the error stream type to `EventStream`.

src/journal/src/grpc/stream.rs
zojw: rust seems lake of [Function_composition](https://wiki.haskell.org/Function_composition), maybe we still need "map nested map", maybe like this but it seems not much better 😞 

```
        Pin::new(&mut self.events).poll_next(cx).map(|o| {
            o.map(|r| match r {
                Ok(resp) => Ok(Event {
                    ts: resp.ts,
                    data: resp.data,
                }),
                Err(s) => Err(Error::from(s)),
            })
        })
```

src/journal/src/mem/stream.rs
huachaohuang: Oh, that's just an example. What I mean is that we can return some stream that simply yields an error immediately.

src/journal/src/grpc/stream.rs
zojw: oh, maybe I know what your means - -....or maybe we can try map in `read_events`~ then we can remove `EventStream` implement code~  I'm not sure it's right but it can pass compiler's check 🤣 

```
diff --git a/src/journal/src/grpc/stream.rs b/src/journal/src/grpc/stream.rs
index 1969f6f..5d1c069 100644
--- a/src/journal/src/grpc/stream.rs
+++ b/src/journal/src/grpc/stream.rs
@@ -17,6 +17,7 @@ use std::{
     task::{Context, Poll},
 };

+use futures::StreamExt;
 use tonic::Streaming;

 use super::{
@@ -41,7 +42,7 @@ impl RemoteStream {
 #[async_trait]
 impl Stream for RemoteStream {
     type Error = Error;
-    type EventStream = EventStream;
+    type EventStream = impl futures::Stream<Item = Result<Event<Self::Timestamp>>> ;
     type Timestamp = u64;

     async fn read_events(&self, ts: Self::Timestamp) -> Result<Self::EventStream> {
@@ -50,7 +51,15 @@ impl Stream for RemoteStream {
             ts,
         };
         let output = self.client.read_event(input).await?;
-        Ok(EventStream::new(output))
+        Ok(output.map( |r| {
+            match r {
+                Ok(resp) => Ok(Event {
+                    ts: resp.ts,
+                    data: resp.data,
+                }),
+                Err(s) => Err(Error::from(s)),
+            }
+        }))
     }

     async fn append_event(&self, event: Event<Self::Timestamp>) -> Result<()> {
diff --git a/src/journal/src/lib.rs b/src/journal/src/lib.rs
index 0df7802..dd06bc4 100644
--- a/src/journal/src/lib.rs
+++ b/src/journal/src/lib.rs
@@ -12,6 +12,8 @@

+#![feature(type_alias_impl_trait)]
+
 mod journal;
 mod stream;

@@ -23,4 +25,4 @@ pub use async_trait::async_trait;
 pub use self::{
     journal::Journal,
     stream::{Event, Stream, Timestamp},
-};
+};
\ No newline at end of file
```

tisonkun:
@huachaohuang @zojw thanks for your comments. I've supported generic Timestamp with `Serialize + DeserializeOwned`.

IMO this patch can be still optimized in how to write `Stream` code and serde code. I'll try to integrate with your comments. And this PR itself is functionally completed. Ready for review.

src/journal/Cargo.toml
```diff
@@ -8,4 +8,12 @@ publish = false
...
+serde_json = "1.0"
```
tisonkun, line 16: Maybe using a more compact format.

src/journal/Cargo.toml
Xuanwo: I'm confused about the `Timestamp`, why we need to have a trait here?

```rust
pub trait Timestamp: Ord + Send + Copy + Debug + Unpin {}

impl<T: Ord + Send + Copy + Debug + Unpin> Timestamp for T {}
```

src/journal/Cargo.toml
Xuanwo: I prefer to use `msgpack`， but I think the format is not the current bottleneck, so we don't need to optimize it at this stage right now.

src/journal/Cargo.toml
huachaohuang: @Xuanwo the discussion is [here](https://github.com/engula/engula/issues/100).

src/journal/Cargo.toml
huachaohuang: It should be as lightweight as possible. But I will accept anything that works for now :)

tisonkun: Yes. Here is an open-ended discussion entrypoint. As long as we don't use nested timestamp actually, it's not quite significant. However, I'd like to learn how the Rust ecosystem does serde XD.

src/journal/src/mem/stream.rs
```diff
@@ -76,6 +76,8 @@ pub struct EventStream<T: Timestamp> {
     offset: usize,
 }
 
+impl<T: Timestamp> Unpin for EventStream<T> {}
```
huachaohuang, line 79: This seems unnecessary.

src/journal/src/grpc/stream.rs
```diff
@@ -0,0 +1,119 @@
...
+        Ok(())
+    }
+}
+
+pub struct EventStream<T: Timestamp> {
+    events: Streaming<ReadEventResponse>,
+    _t: PhantomData<T>,
+}
+
+impl<T: Timestamp> Unpin for EventStream<T> {}
```
huachaohuang, line 90: ditto

src/journal/src/grpc/stream.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+impl<T: Timestamp> futures::Stream for EventStream<T> {
+    type Item = Result<Event<T>>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match Pin::new(&mut self.events).poll_next(cx) {
+            Poll::Ready(Some(Ok(resp))) => Poll::Ready(Some(Ok(Event {
+                ts: deserialize_ts(resp.ts.as_slice())?,
```
huachaohuang, line 107: Does `&resp.ts` work here?

huachaohuang (commented):
Looks good to me overall. The `serde` things looks a bit weird. But I will accept it as long as it works for now :)

src/journal/src/mem/stream.rs
tisonkun: Nice catch! Updated.

src/journal/src/grpc/stream.rs
tisonkun: Yes. I'll update it. Rust provides both symbol & method to do the same thing, interesting 🤔

huachaohuang: `&resp.ts` can be inferred to `&Vec<u8>` or `&[u8]`. I think there are some cases where the compiler can't tell which one to use. Then you need to use `as_slice()` instead.

tisonkun: Updated.

src/journal/src/mem/stream.rs
tisonkun: I try to let EventStream becomes an enum that can be either a current EventStream or a wrapper of error to be yield when calling `poll_next`, but match on a pinning enum is quite tricky :(

I didn't find a good way to implement this "either" logic.

src/journal/src/grpc/stream.rs
```diff
@@ -0,0 +1,117 @@
...
+        Ok(EventStream::new(output))
+    }
+
+    async fn append_event(&self, event: Event<Self::Timestamp>) -> Result<()> {
+        let input = AppendEventRequest {
+            stream: self.stream.clone(),
+            ts: serialize_ts(&event.ts)?,
+            data: event.data,
+        };
+        let _ = self.client.append_event(input).await?;
```
huachaohuang, line 71: Is `let _ = ` necessary here?

src/journal/src/grpc/stream.rs
```diff
@@ -0,0 +1,117 @@
...
+        Ok(())
+    }
+
+    async fn release_events(&self, ts: Self::Timestamp) -> Result<()> {
+        let input = ReleaseEventsRequest {
+            stream: self.stream.clone(),
+            ts: serialize_ts(&ts)?,
+        };
+        let _ = self.client.release_events(input).await?;
```
huachaohuang, line 80: ditto

src/journal/src/grpc/journal.rs
```diff
@@ -0,0 +1,61 @@
...
+impl<T: Timestamp> Journal<RemoteStream<T>> for RemoteJournal<T> {
+    async fn stream(&self, name: &str) -> Result<RemoteStream<T>> {
+        Ok(RemoteStream::new(self.client.clone(), name.to_owned()))
+    }
+
+    async fn create_stream(&self, name: &str) -> Result<RemoteStream<T>> {
+        let input = CreateStreamRequest {
+            stream: name.to_owned(),
+        };
+        let _ = self.client.create_stream(input).await?;
```
huachaohuang, line 50: ditto

src/journal/src/grpc/journal.rs
```diff
@@ -0,0 +1,61 @@
...
+        self.stream(name).await
+    }
+
+    async fn delete_stream(&self, name: &str) -> Result<()> {
+        let input = DeleteStreamRequest {
+            stream: name.to_owned(),
+        };
+        let _ = self.client.delete_stream(input).await?;
```
huachaohuang, line 58: ditto

huachaohuang (commented):
Rest LGTM

huachaohuang:
Maybe the name "stream" for Journal is a bad choice. Because when it is a bit confused when mixed with the `Stream` trait in Rust.

src/journal/src/grpc/stream.rs
tisonkun: It's unnecessary. Thanks for spotting it. Remove.

huachaohuang (approved):
LGTM, thanks for digging :)

PR #122, opened 2021-11-19
journal: replace Result<Stream> to an either stream

tisonkun:

This is a follow-up to #119 and thus ref #117 .


huachaohuang:
@tisonkun thanks for exploring. But maybe we should keep the current way, considering the extra complexity and inefficiency (doing a match in every `next`).

tisonkun:
@huachaohuang fine.

Closed as invalid.

PR #124, opened 2021-11-20
ci: integration tests for s3

### LINKED ISSUE #109: Investigate how to test with s3
ref https://github.com/engula/engula/pull/99#discussion_r748133615

need some method to automatic test with AWS service like s3 in CI without leaking access&secret key, IMHO, maybe we can do it in two directions:

- unit test: mock AWS SDK, there are exists an [issue](https://github.com/awslabs/aws-sdk-rust/issues/199#issuecomment-904558631 ) indicates SDK support TestConnection to mock/record request-response and we take some [tries](https://github.com/engula/engula/pull/99/files#diff-f6197926e0b89ec759315b9bd0efe5800b0eb88a71ece6d2a60746dd07b98af7R60), but it seems we need more encapsulation works to make it easy for use.
- integration-test: deploy a temp S3-like service(for example https://min.io/) in CI pipeline and make test case using temp S3-like service.

Welcome to give more advice or contribution code 😄

tisonkun:
Commits are formatted. This closes #109 .


src/platform/aws/tests/s3_storage.rs
```diff
@@ -0,0 +1,90 @@
...
+use aws_types::{region::Region, Credentials};
+use http::Uri;
+use smithy_http::endpoint::Endpoint;
+
+const ACCESS_KEY: &str = "engulatest";
+const SECRET_KEY: &str = "engulatest";
+const TEST_REGION: &str = "us-east-2";
+
+#[tokio::test]
+#[ignore]
```
tisonkun, line 27: covered by:

```yml
    - name: Cargo test
      uses: actions-rs/cargo@v1
      with:
        command: test
        args: --manifest-path src/platform/aws/Cargo.toml -- --ignored
```

I'm looking for better tests grouped (enabled/disabled) mechanism.

tisonkun: Although I found a convention to name all IT with `integration_test_` prefix, now I tend to `#[ignore]` all expensive tests so that developers can easily run all UT with `cargo test --workspace` without further knowledge.

In my investigation, if every IT with `integration_test_` prefix and without `#[ignore]`, a developer can:

* run UT with: `cargo test --workspace -- --skip integration_test_`
* run specific IT with: `cargo test <...> integration_test_`

huachaohuang: Nice

src/platform/aws/tests/s3_storage.rs
```diff
@@ -0,0 +1,90 @@
...
+    let config = Config::builder()
+        .region(region)
+        .credentials_provider(credentials)
+        .endpoint_resolver(endpoint)
+        .build();
+
+    let storage = S3Storage::new(TEST_REGION, config);
+
+    let bucket = "tests-multipart-put";
+    let object = "test-object-0";
```
huachaohuang, line 74: Forgive my obsession, this misses an "s". Or maybe drop the "tests", it's cleaner.

huachaohuang (commented):
Rest LGTM. @zojw do you want to take a look?

.github/workflows/audit-license.yml
```diff
@@ -16,6 +16,11 @@ name: Audit License
 
 on: [push, pull_request]
 
+# See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency.
```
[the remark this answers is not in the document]
tisonkun: Yep. Removing...

.github/workflows/audit-license.yml
```diff
@@ -16,6 +16,11 @@ name: Audit License
...
+concurrency:
+  group: ${{ github.ref }}-${{ github.workflow }}
```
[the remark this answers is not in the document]
tisonkun: Yes. It can be interpreted as "cancel previous running instance" (because a new commit is pushed). In this way, we save some less interesting consumptions for CI resources.

.github/workflows/s3.yml
```diff
@@ -0,0 +1,74 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: S3 Integration Tests
+
+on:
+  push:
+    path:
+      - 'platform/aws/'
```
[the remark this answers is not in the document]
tisonkun: Yes. I think running S3 specific IT for all is unnecessary and writing this logic.

I know that basic logic outside the directory can affect IT but we can easily modify a dummy file under the path to trigger this test.

However, given the actual workflow takes about 5 minutes to run, I'm OK to use the same rule as checkcode.yml.

What do you think @huachaohuang ?

src/platform/aws/tests/s3_storage.rs
tisonkun: Make sense. Dropping all `test(s)-` prefixes for these names...

.github/workflows/s3.yml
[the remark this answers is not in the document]
huachaohuang: I think the current setting is perfect. Let's keep it as it is.

huachaohuang (approved):
LGTM

PR #125, opened 2021-11-20
storage: add local file storage

zojw:
ref https://github.com/engula/engula/issues/68#issuecomment-972728919

implement simple local file storage, using `tokio::fs`'s API


src/storage/src/local/storage.rs
```diff
@@ -0,0 +1,79 @@
...
+    object::LocalObject,
+};
+use crate::{async_trait, Storage};
+
+pub struct LocalStorage<'a> {
+    root: Cow<'a, Path>,
+}
+
+impl<'a> LocalStorage<'a> {
+    pub async fn from(root: impl Into<Cow<'a, Path>>) -> Result<LocalStorage<'a>> {
```
huachaohuang, line 34: I think `new` is better, because `from` is usually used in type conversion. And do we really need `Into<Cow<'a, Path>>` here? I think `AsRef<Path>` is more common and simpler.

src/storage/src/local/storage.rs
```diff
@@ -0,0 +1,79 @@
...
+        let path = root.into();
+        fs::DirBuilder::new()
+            .recursive(true)
+            .create(path.as_ref())
+            .await?;
+        Ok(Self { root: path })
+    }
+
+    fn bucket_path(&self, name: impl AsRef<Path>) -> PathBuf {
+        self.root.as_ref().join(name)
```
huachaohuang, line 44:
```suggestion
        self.root.join(name)
```

src/storage/src/local/storage.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        fs::create_dir_all(path.as_path()).await?;
+
+        Ok(LocalBucket::new(path))
+    }
+
+    async fn delete_bucket(&self, name: &str) -> Result<()> {
+        let path = self.bucket_path(name);
+
+        fs::remove_dir(path.as_path()).await?;
```
huachaohuang, line 75:
```suggestion
        fs::remove_dir(path).await?;
```

huachaohuang: I see you convert `Path` to `PathBuf` in some places. It is not necessary in most cases, because these functions accept `AsRef<Path>`, which applies to `Path` and `PathBuf`.

src/storage/src/local/object.rs
```diff
@@ -0,0 +1,60 @@
...
+    async fn read_at(&self, mut buf: &mut [u8], offset: usize) -> Result<usize> {
+        let mut f: fs::File = fs::OpenOptions::new()
+            .read(true)
+            .open(self.path.as_ref())
+            .await?;
+
+        f.seek(SeekFrom::Start(offset as u64)).await?;
+
+        let mut read_size: usize = 0;
+        while !buf.is_empty() {
```
huachaohuang, line 48: I think [`read_exact`](https://docs.rs/tokio/1.14.0/tokio/io/trait.AsyncReadExt.html#method.read_exact) is what you need.

src/storage/src/local/bucket.rs
```diff
@@ -0,0 +1,103 @@
...
+    async fn finish(self) -> Result<usize> {
+        let mut f = fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(self.path.as_ref())
+            .await?;
+
+        let v: &[u8] = &self.buf;
+        f.write_all(v).await?;
```
huachaohuang, line 98:
```suggestion
        f.write_all(&self.buf).await?;
```

src/storage/src/local/storage.rs
```diff
@@ -0,0 +1,79 @@
...
+            return Err(Error::NotFound(name.to_owned()));
+        }
+
+        Ok(LocalBucket::new(path))
+    }
+
+    async fn create_bucket(&self, name: &str) -> Result<LocalBucket<'a>> {
+        let path = self.bucket_path(name);
+
+        if fs::metadata(path.as_path()).await.is_ok() {
```
huachaohuang, line 63:
```suggestion
        if fs::metadata(&path).await.is_ok() {
```

src/storage/src/lib.rs
```diff
@@ -17,6 +17,7 @@ mod object;
 mod storage;
 
 pub mod grpc;
+pub mod local;
```
huachaohuang, line 20: How about `file`, "file storage" analogy with "file system". "Local" is a broader concept, for example, the "mem" implementation can be regarded as "local" too.

src/storage/src/local/object.rs
zojw: read_exact will return an error when it encounters an “end of file” before completely filling the buffer, so this case will be failure https://github.com/engula/engula/pull/125/files#diff-8dbf6cd86e09442d3c36fb41406a11a315afb695188f842b5d04fa64b2697eb9R132 (read with 4 bytes buf but only have 3 bytes data), IMO, we should return 3 and fill 3 bytes in the buffer~

src/storage/src/local/storage.rs
zojw: emm~ yes, it seems Cow isn't very suitable for here.... and compare with AsRef maybe Into<PathBuf> will more suitable for this, because storage should own path, so we need `.as_ref().to_owned()` to save root path but will cause addition allocation when we pass in a PathBuf argument

src/storage/src/file/bucket.rs
```diff
@@ -0,0 +1,99 @@
...
+        Ok(())
+    }
+}
+
+pub struct FileObjectUploader {
+    path: PathBuf,
+    buf: Vec<u8>,
+}
+
+impl<'a> FileObjectUploader {
```
huachaohuang, line 68: This `<'a>' seems unused.

src/storage/src/file/bucket.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn new(path: impl Into<PathBuf>) -> Self {
+        Self {
+            path: path.into(),
+            buf: vec![],
+        }
+    }
+}
+
+#[async_trait]
+impl<'a> ObjectUploader for FileObjectUploader {
```
huachaohuang, line 78: ditto

src/storage/src/file/bucket.rs
```diff
@@ -0,0 +1,99 @@
...
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(&self.path)
+            .await?;
+
+        f.write_all(&self.buf).await?;
+        f.sync_all().await?;
+
+        Ok(1)
```
huachaohuang, line 97: Shouldn't we return the buffer size?

src/storage/src/file/bucket.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+#[async_trait]
+impl Bucket<FileObject> for FileBucket {
+    type ObjectUploader = FileObjectUploader;
+
+    async fn object(&self, name: &str) -> Result<FileObject> {
+        let path = self.object_path(name);
+        if fs::metadata(&path).await.is_err() {
```
huachaohuang, line 45: I think a lot of errors can occur here. It may not be appropriate to map all errors to `NotFound`.

src/storage/src/file/storage.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+#[async_trait]
+impl Storage<FileObject, FileBucket> for FileStorage {
+    async fn bucket(&self, name: &str) -> Result<FileBucket> {
+        let path = self.bucket_path(name);
+
+        if fs::metadata(&path).await.is_err() {
+            return Err(Error::NotFound(name.to_owned()));
```
huachaohuang, line 48: ditto

src/storage/src/file/object.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        let mut read_size: usize = 0;
+        while !buf.is_empty() {
+            let n = f.read(buf).await?;
+            if n == 0 {
+                break;
+            }
+            read_size += n;
+            let tmp = buf;
+            buf = &mut tmp[n..];
```
huachaohuang, line 52:
```suggestion
            buf = &mut buf[n..];
```

src/storage/src/file/bucket.rs
zojw: oops...my fatal - -

src/storage/src/file/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+
+#[cfg(test)]
+mod tests {
+    use std::{env, path::Path, process};
+
+    use super::{error::Result, *};
+    use crate::*;
+
+    struct TestEnvGuard<'a> {
+        path: Cow<'a, Path>,
```
huachaohuang, line 32: How about `PathBuf`?

src/storage/src/file/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+        u.finish().await?;
+        let mut u = b.upload_object("obj-1").await?;
+        u.write(b"123").await?;
+        u.finish().await?;
+
+        let obj = b.object("obj-1").await?;
+        let mut v = [0u8; 4];
+        let n = obj.read_at(&mut v[..], 0).await?;
+        assert_eq!(n, 3);
+        assert_eq!(&v[..n], "123".as_bytes());
```
huachaohuang, line 134:
```suggestion
        assert_eq!(&v[..n], b"123");
```

src/storage/src/file/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        let mut v = [0u8; 2];
+        let n = obj.read_at(&mut v[..], 0).await?;
+        assert_eq!(n, 2);
+        assert_eq!(&v[..n], "12".as_bytes());
```
huachaohuang, line 139:
```suggestion
        assert_eq!(&v[..n], b"12");
```

huachaohuang (commented):
Rest LGTM

huachaohuang (approved):
LGTM, thanks!

src/storage/src/file/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+    use crate::*;
+
+    struct TestEnvGuard {
+        path: PathBuf,
+    }
+
+    impl TestEnvGuard {
+        fn setup(case: &str) -> Self {
+            let mut path = env::temp_dir();
+            path.push("englua-storage-fs-test");
```
tisonkun, line 38:
```diff
diff --git a/src/storage/src/file/mod.rs b/src/storage/src/file/mod.rs
index 4832b0c..caa0ab0 100644
--- a/src/storage/src/file/mod.rs
+++ b/src/storage/src/file/mod.rs
@@ -35,7 +35,7 @@ mod tests {
     impl TestEnvGuard {
         fn setup(case: &str) -> Self {
             let mut path = env::temp_dir();
-            path.push("englua-storage-fs-test");
+            path.push("engula-storage-fs-test");
             path.push(process::id().to_string());
             path.push(case);
             Self { path }
```

you may consider fix it at #129.

huachaohuang: Wow, you have a very good eye 😆 . Maybe simply drop those unnecessary prefixes.

tisonkun: Agree. It's just arbitary magic literal :P

PR #126, opened 2021-11-21
Add editorconfig

### LINKED ISSUE #120: Find tool and add Github Action to check toml files
We have two toml files style now. For example, [this](https://github.com/engula/engula/blob/main/src/platform/aws/Cargo.toml) uses the style:

```
tokio = {version = "1.13", features = ["full"]}
```

But [this](https://github.com/engula/engula/blob/main/src/storage/Cargo.toml) use the style:

```
tokio = { version = "1.13", features = ["full"] }
```

Note the whitespace between `{}`. It would be nice to keep toml files consistent for all contributors.

tisonkun:
cc @huachaohuang 

I don't know whether we should introduce a CI task. Many projects only add this config file and most editors (vscode, jetbrains, vim, emacs, ...) support applying editorconfig by plugins already. If we introduce a CI task, we trade off overhead for strict requirements. Otherwise, it's a hint/auxiliary and loosely guarded by reviewers.


PsiACE:
Just add some hints. I think this is not a necessary CI check. Most of the time, cargo fmt is sufficient.

tisonkun:
@PsiACE I agree. But firstly the community should agree on the rules and later perform during reviews.

huachaohuang:
Nice. I think we can keep the CI task until people start complaining about it :)
But we still have the problem that there are two toml files style now. Check [this](https://github.com/engula/engula/blob/main/src/platform/aws/Cargo.toml) and [this](https://github.com/engula/engula/blob/main/src/storage/Cargo.toml).

tisonkun:
@huachaohuang OK. Then I remove the "close" statement. I focus on the irrelevant blank though.

For toml file "style", could you elaborate a bit on the issue of how they are different? Also, I guess there is no checker for toml style yet. We can write a style guide for it and apply it during reviewing.

huachaohuang:
> For toml file "style", could you elaborate a bit on the issue of how they are different? Also, I guess there is no checker for toml style yet. We can write a style guide for it and apply it during reviewing.

https://github.com/engula/engula/issues/120 updated.

huachaohuang (approved):

PR #129, opened 2021-11-22
storage: merge bucket trait into storage trait

### LINKED ISSUE #127: Merge Bucket into Storage
We have three traits for the storage component now: `Storage`, `Bucket`, `Object`. This brings two problems:

- The trait bound is a bit complicated, like `Storage<Object, Bucket>`
- There are some corner cases dealing with individual buckets, for example, https://github.com/engula/engula/pull/86#discussion_r750893023

A simpler solution is to merge Bucket into Storage, so Storage will provide the following interfaces:

- list buckets
- create bucket
- delete bucket
- list objects in a bucket
- upload an object to a bucket
- delete an object from a bucket

We can still keep `Object` and `ObjectUploader` as usually. Note that only `Object` is performance-critical here, as users may read from an object frequently. Other storage operations don't need to care about performance very much.

zojw:
closes #127

merge bucket trait into storage trait, there are mainly 4 commits to help review:

- merge API and fix error in storage crate
- fix error in platform AWS trait
- rename objectUploader filename
- fix broken bucket exist check in storage/fs impl


src/storage/src/file/mod.rs
```diff
@@ -92,7 +90,7 @@ mod tests {
         let r = s.create_bucket(BUCKET_NAME).await;
         assert!(r.is_err());
         assert_eq!(
-            "`test_bucket_dup` already exists",
+            "`bucket 'test_bucket_dup'` already exists",
             r.err().unwrap().to_string()
         );
```
huachaohuang, line 95: You can try `assert!(matches!(r, Err(Error::AlreadyExists(_))))`

src/storage/src/file/mod.rs
```diff
@@ -139,4 +137,36 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_not_exist_bucket() -> Result<()> {
+        const BUCKET_NAME: &str = "test_not_exist_bucket";
+        let g = TestEnvGuard::setup("test_not_exist_bucket");
+        let s = FileStorage::new(&g.path).await?;
+
+        let r = s.delete_object(BUCKET_NAME, "obj-1").await;
+        assert!(r.is_err());
+        assert_eq!(
+            "`bucket 'test_not_exist_bucket'` is not found",
```
huachaohuang, line 150: ditto

src/storage/src/mem/storage.rs
```diff
@@ -64,4 +71,49 @@ impl Storage<MemObject, MemBucket> for MemStorage {
...
+            Some(_) => Ok(()),
+            None => Err(Error::NotFound(format!("object '{}'", object_name))),
+        }
+    }
+}
+
+pub type Objects = Arc<Mutex<HashMap<String, MemObject>>>;
+
+#[derive(Clone)]
+struct MemBucket {
```
huachaohuang, line 109: Shouldn't we merge the bucket into the storage? You know, to prevent [this case](https://github.com/engula/engula/pull/86#discussion_r750893023)

src/storage/src/storage.rs
```diff
@@ -12,17 +12,28 @@
...
 #[async_trait]
-pub trait Storage<O: Object, B: Bucket<O>> {
-    /// Returns a bucket.
-    async fn bucket(&self, name: &str) -> Result<B, O::Error>;
-
+pub trait Storage<O: Object> {
+    type ObjectUploader: ObjectUploader<Error = O::Error>;
     /// Creates a bucket.
-    async fn create_bucket(&self, name: &str) -> Result<B, O::Error>;
+    async fn create_bucket(&self, name: &str) -> Result<(), O::Error>;
```
huachaohuang, line 22: Maybe we should change these interfaces to accept `<S: Into<String>>`, because I see most implementations require the ownership of the name.

huachaohuang (commented):
It looks great! Leave some comments for now, will check again later.

src/storage/src/storage.rs
zojw: I had try it in 57576639827d3b105ca867b2f4939f9507802371, but it seems not pretty due to `+ Send + lifetime`, I try to use marco but I'm not sure is it practice to do that- -

huachaohuang (approved):
LGTM, I think the storage component looks much better now :)

Issue #58, opened 2021-11-02
Roadmap v0.2 - Microunit


huachaohuang:
Non-goals:

- Data persistence
- Leader election and failover
- Resource isolation and management

Tasks:

- [ ] Design document
- [ ] Commands
  - [ ] Start a node
  - [ ] Bootstrap a universe
  - [ ] Join a node to a universe

huachaohuang:
We need to decide the command line interface. The [Command Line Interface Guidelines](https://clig.dev/) is awesome. I prefer the `noun verb` style, like `docker container create`.

So for example, we can have the following subcommands:

- `engula node` to operate on a specific node
- `engula unit` to operate on a specific unit

tisonkun:
> We need to decide the command line interface. The [Command Line Interface Guidelines](https://clig.dev/) is awesome. I prefer the `noun verb` style, like `docker container create`.
> 
> So for example, we can have the following subcommands:
> 
> * `engula node` to operate on a specific node
> * `engula unit` to operate on a specific unit

The guidelines look awesome! Shall we name it `engula`? I can see a `engula` crate (what does it for) so far and am afraid of the word gets override too much.

But it's OK we implement a ctl of microunit and disptach `engula node` to `microunit node` or `muctl node` as a facade.

tisonkun:
>  I think we don't need a consistent store for v0.2. We can store all data in memory for now.

Does it mean we deliver a single-node version microunit in v0.2? Or we can use a preconfigured leader election service and thus all nodes know where to get the data / register themselves?

tisonkun:
> @huachaohuang Node should provide HTTP API ...

I'd like to know whether we will make use of a certain HTTP server library or write from what Rust stdlib provides?

huachaohuang:
> > We need to decide the command line interface. The [Command Line Interface Guidelines](https://clig.dev/) is awesome. I prefer the `noun verb` style, like `docker container create`.
> > So for example, we can have the following subcommands:
> > 
> > * `engula node` to operate on a specific node
> > * `engula unit` to operate on a specific unit
> 
> The guidelines look awesome! Shall we name it `engula`? I can see a `engula` crate (what does it for) so far and am afraid of the word gets override too much.
> 
> But it's OK we implement a ctl of microunit and disptach `engula node` to `microunit node` or `muctl node` as a facade.

Actually, the binary does not belong to microunit, it is a tool for the whole Engula project. Maybe we should move the discussion somewhere else.

[...]

PR #130, opened 2021-11-22
microunit: add control unit

huachaohuang:
This PR implements some ideas from https://github.com/engula/engula/issues/58#issuecomment-974195015 and https://github.com/engula/engula/issues/58#issuecomment-974831557.
It also implements [`IntoResponse`](https://docs.rs/axum/0.3.4/axum/response/trait.IntoResponse.html) for `Error` to convert errors to responses automatically.

We can create a control unit now. First of all, start a node:

```
> cargo run -p engula -- node start 127.0.0.1:1234
```

Then create a control unit:

```
> curl --header "Content-Type: application/json" \
--request POST \
--data '{"kind": "ControlUnit"}' \
http://127.0.0.1:1234/v1/units
```

List all units to get the address of the control unit:

```
> curl http://127.0.0.1:1234/v1/units             
[{"id":"bf099b3f-e6ba-4160-96bf-340dcf7bd048","kind":"ControlUnit","addr":"0.0.0.0:50830"}]
```

Now we can query the control unit:

```
> curl http://127.0.0.1:50830/v1/universe/nodes
[]
```

### DIFF
--- a/src/microunit/Cargo.toml
+++ b/src/microunit/Cargo.toml
@@ -7,6 +7,7 @@ publish = false
 [dependencies]
 async-trait = "0.1"
 axum = "0.3"
+hyper = "0.14"
 tokio = "1.13"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 serde = { version = "1.0", features = ["derive"] }
--- a/src/microunit/Cargo.toml
+++ b/src/microunit/Cargo.toml
@@ -13,3 +13,5 @@ uuid = { version = "0.8", features = ["serde", "v4"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 thiserror = "1.0"
+reqwest = { version = "0.11", features = ["json"] }
+url = "2.2"
--- a/src/microunit/Cargo.toml
+++ b/src/microunit/Cargo.toml
@@ -11,3 +11,4 @@ tokio = "1.13"
 uuid = { version = "0.8", features = ["serde", "v4"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+thiserror = "1.0"
--- a/src/microunit/src/control_client.rs
+++ b/src/microunit/src/control_client.rs
@@ -12,37 +12,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
+use std::net::SocketAddr;
 
-use tokio::sync::Mutex;
+use reqwest::{Client, Url};
 
 use crate::{error::Result, proto::*};
 
-pub struct Control {
-    inner: Mutex<Inner>,
+pub struct ControlClient {
+    url: Url,
+    client: Client,
 }
 
-impl Default for Control {
-    fn default() -> Self {
-        let inner = Inner::default();
-        Self {
-            inner: Mutex::new(inner),
-        }
-    }
-}
-
-#[derive(Default)]
-struct Inner {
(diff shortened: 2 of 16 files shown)


zojw:
>Now we can query the control unit:
>> curl http://127.0.0.1:50830/v1/universe/nodes
>>[]

maybe the `--initial` node needs immediately auto-join itself to control-unit after created~? if we choose control-unit pull node way as https://github.com/engula/engula/issues/58#issuecomment-974195015

src/microunit/src/control_unit.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#[async_trait]
+impl Unit for ControlUnit {
+    async fn desc(&self) -> UnitDesc {
+        self.desc.lock().await.clone()
+    }
+
+    async fn start(&self) -> UnitResult<()> {
+        let addr = ([0, 0, 0, 0], 0).into();
```
zojw, line 52: LGTM for me for the initial version~ but I have some questions about the final version.. 😄 

1. current version we only can get 0.0.0.0, but maybe we need some way to get accessible IP for control_unit, for example, add a [`advertise-xx`]( https://etcd.io/docs/v3.3/faq/#what-is-the-difference-between-listen-clientpeer-urls-advertise-client-urls-or-initial-advertise-peer-urls) params to node-ctl
2. do we choose to open one port for each unit, maybe it will have some network policy question in some strict env

src/microunit/Cargo.toml
```diff
@@ -7,7 +7,11 @@ publish = false
 [dependencies]
 async-trait = "0.1"
 axum = "0.3"
+hyper = "0.14"
```
tisonkun, line 10: It seems this dep is unused.

src/microunit/src/control_client.rs
```diff
@@ -0,0 +1,38 @@
+// Copyright 2021 The Engula Authors.
```
tisonkun, line 1: deadcode?

tisonkun: At least it's not exposed in `lib.rs` in contrast to `NodeClient`

huachaohuang:
> maybe the `--initial` node needs immediately auto-join itself to control-unit after created~? if we choose control-unit pull node way as [#58 (comment)](https://github.com/engula/engula/issues/58#issuecomment-974195015)

Yes, that's the plan. I will do it in the next few PRs.

tisonkun:
This implementation is different from:

> So if a node receives requests against `/v1/universe`, it will redirect the client to the control plane. In this case, users can send universe requests to any node and any control unit.

as it mounts API for node in a different port from that for unit.

huachaohuang:
> This implementation is different from:
> 
> > So if a node receives requests against `/v1/universe`, it will redirect the client to the control plane. In this case, users can send universe requests to any node and any control unit.
> 
> as it mounts API for node in a different port from that for unit. And if you request the node for unit API, it won't be found.
> 
> ```
> $ curl -i http://127.0.0.1:1234/v1/universe/nodes
> HTTP/1.1 404 Not Found
> content-length: 0
> date: Tue, 23 Nov 2021 12:20:06 GMT
> ```

Yeah, that's because I haven't implemented the redirection part.

huachaohuang:
Since we are shifting to k8s, I think microunit will be redesigned and rewritten sooner or later. As for this PR, we can simply merge it to preserve the history or close it. @zojw @tisonkun What do you think?

tisonkun:
@huachaohuang you can simply merge it as a better start point than "hello_unit" :)

PR #131, opened 2021-11-23
Update the architecture image

huachaohuang:
Only the image is updated for now. Descriptions will be updated in the future.

Architecture


huachaohuang:
Since this is not harmful, I will simply merge it now. Feel free to leave comments, though.

PR #136, opened 2021-11-27
kernel: add some basic structs

huachaohuang:
WIP, send a PR first in case someone wants to know what `Kernel` looks like :)


tisonkun (commented):
How can an `Engine` be generic? Doesn't it always take dynamic Storage/Journal and apply requests? What's a remote `Engine` or other implementation?

If we can reduce the generic composition, the complexity will be significantly lowered.

huachaohuang:
@tisonkun I can see two reasonable implementation so far: local and remote. The local implementation simply talks to the underlying journal/storage to do its job. The remote implementation talks to a remote kernel service that manages other things for it.

tisonkun:
@huachaohuang let me list concepts we define so far:

1. Journal, Storage will have multiple implementations, because they can be outside systems. A RemoteJournal makes sense because someone possibly provides their own Journal service.
2. Kernel, if it means resource management, will have multiple implementations in the dimension of resource providers (standalone, YARN, Kubernetes). Because different resource providers have different interaction logic. Any kernel should be able to start any implementation of Journal/Storage, with its specific resource pool.
3. Engine, which talks to Journal/Storage that are internal concepts of Engula, should be a sole implementation, no matter what implementation of Journal/Storage or Engine it will talk to.

huachaohuang:
Close this in favor of #155

PR #137, opened 2021-11-28
storage: refactor with trait objects to allow composition

huachaohuang:
Upper-level modules like Kernel/Engine need to create a storage dynamically without knowing the specific type. Check [this](https://github.com/engula/engula/pull/136) for an example.

Other updates in this PR:

- Update package name to address [this issue](https://github.com/engula/engula/issues/59#issuecomment-980544605)
- Update toolchain to address [this issue](https://github.com/rust-analyzer/rust-analyzer/issues/10772).

I am currently focusing on prototyping the [kernel[(https://github.com/engula/engula/pull/136). So I opt out of some implementations to avoid fixing all of them now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,147 +74,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
-name = "aws"
-version = "0.2.0"
-dependencies = [
- "aws-endpoint",
- "aws-sdk-s3",
- "aws-types",
- "bytes",
- "http",
- "smithy-http",
- "storage",
- "thiserror",
- "tokio",
-]
-
-[[package]]
-name = "aws-auth"
-version = "0.1.0"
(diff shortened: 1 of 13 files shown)


tisonkun (dismissed):
If we're re-design the abstraction, it's totally OK to remove the outdated implementations and bring them back while the abstraction becomes solid again. One of the advantages of using VCS is to develop rapidly and clearly while keeping all the version history.

There is always an overhead to review the following PR to bring back/update other implementations (as #99). If we're in a situation to prototyping with mem impl rapidly, we can just frankly reflect it with the main branch :)

For the diff, it looks good to me that we use a `Box<dyn Trait>` for prototyping and later optimizing with enum dispatch or something.

Please fix the unnecessary deps warning or simply disable/remove it for now.

tisonkun:
Otherwise, you can prototype the kernel concept in your fork, and bring changes only if it seems a good direction. You should lose nothing but like developing on the main branch.

huachaohuang:
@tisonkun Yeah, maybe just keep these PRs open in case someone interested in what we are doing.

tisonkun:
@huachaohuang Make sense. You can convert it into a draft or set a highlight note on the description. I'm also digging a better abstraction.

huachaohuang:
Close in favor of #154 .

PR #138, opened 2021-11-28
journal: refactor with trait objects for composition

huachaohuang:
Check https://github.com/engula/engula/pull/137 for more details.

### DIFF
--- a/src/journal/src/journal.rs
+++ b/src/journal/src/journal.rs
@@ -12,16 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{async_trait, Result, Stream, Timestamp};
+use crate::{async_trait, Result, Stream};
 
 /// An interface to manipulate a journal.
 #[async_trait]
-pub trait Journal<T: Timestamp> {
+pub trait Journal: Send + Sync {
     /// Returns a stream.
-    async fn stream(&self, name: &str) -> Result<Box<dyn Stream<T>>>;
+    async fn stream(&self, name: &str) -> Result<Box<dyn Stream>>;
 
     /// Creates a stream.
-    async fn create_stream(&self, name: &str) -> Result<Box<dyn Stream<T>>>;
+    async fn create_stream(&self, name: &str) -> Result<Box<dyn Stream>>;
 
     /// Deletes a stream.
     async fn delete_stream(&self, name: &str) -> Result<()>;
--- a/src/journal/src/journal.rs
+++ b/src/journal/src/journal.rs
@@ -12,17 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::{async_trait, stream::Stream};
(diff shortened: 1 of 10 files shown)


huachaohuang:
The generic timestamp draws me crazy in https://github.com/engula/engula/pull/136. We should consider removing it. Maybe users can embed their timestamps in the data somehow when necessary.

tisonkun:
> The generic timestamp

@huachaohuang if it turns out to be premature optimization, we can inline it as `u64` and make it generic when actually need it. We shouldn't block ourselves by this stuff.

tisonkun:
@huachaohuang I create PR #139 to avoid shooting our feet with premature optimization `Timestamp`.

Also, flattening the `Error` type seems a good direction to reduce generic composition. And I'm digging how to avoid the `EventStream` type parameter.

src/journal/src/journal.rs
```diff
@@ -12,17 +12,17 @@
...
 
-use super::{async_trait, stream::Stream};
+use crate::{async_trait, Result, Stream};
 
 /// An interface to manipulate a journal.
 #[async_trait]
-pub trait Journal<S: Stream> {
+pub trait Journal: Send + Sync {
     /// Returns a stream.
-    async fn stream(&self, name: &str) -> Result<S, S::Error>;
```
tisonkun, line 21: The problem of `S::Error` here is about concepts leak: Why does a user of Journal should be aware of concrete errors at stream level and even distinguished by implementation? How can he/she handle it? If the error cannot be resolved at the low level, it's normally cannot be resolved at the high level. At least when resolving at the high level, the concrete errors is redundant.

I suggest two candidates overcome this problem:

* General `(error_code, error_msg)` structure for all errors in Engula. The downside is that we cannot ensure the error is handled exhausted but leave the responsibility to users handling proper errors. Backtrace can be logged instead of propagated.
* Return proper error type at each level, and propagate as little as possible. Thus, a user of the Journal interface isn't aware of Stream Error or even concrete error of a specific journal impl. We define different errors for different layers. This will increase the impl overhead a lot. Two extremely simple impls are returning `anyhow::Result` at every level and flattening all errors into one enum.

I learn from several projects, most of which define little customized errors and shadow low-level errors quickly.

tisonkun: Always returning `anyhow::Result` is a good start point for prototyping. You can still downcast or attach context. When we actually want to restrict error possibility at some level, we can push more details. I prefer this way.

Adopting `(error_code, error_msg)` is a design decision. Trade off context/error detailed information for easy implementation.

huachaohuang: Yes, we can only expose a few common errors that make sense to users. I am going in that direction now, you can see how the kernel handle errors [here](https://github.com/engula/engula/pull/136/files#diff-2d2d5abd531f2b56312f913802950574f9de8cb5738d5aca8684762e9bbecc6e)

huachaohuang:
Close this in favor of #153

PR #139, opened 2021-11-29
journal: specialize Timestamp

tisonkun:
Change Timestamp from trait to a newtype wrapper of u64.

It turns out that a trait-bound marker of Timestamp goes into premature optimization. Even a most complex clock consists of no more than three parts (seconds, nanoseconds, logical clock, all of which are integers).

We can later extend the Timestamp structure to simulate multiple parts, but specialize & simplify it for now.

This is a follow-up to #103.
This refers to #100.

cc @huachaohuang @Xuanwo

### DIFF
--- a/src/journal/src/stream.rs
+++ b/src/journal/src/stream.rs
@@ -14,21 +14,22 @@
 
 use std::fmt::Debug;
 
-use serde::{de::DeserializeOwned, Serialize};
+use serde::{Deserialize, Serialize};
 
 use super::async_trait;
 
-/// A generic timestamp to order events.
-pub trait Timestamp:
-    Ord + Send + Sync + Copy + Debug + Unpin + Serialize + DeserializeOwned
-{
+#[derive(PartialEq, PartialOrd, Clone, Copy, Debug, Serialize, Deserialize)]
+pub struct Timestamp(u64);
+
+impl From<u64> for Timestamp {
+    fn from(t: u64) -> Self {
+        Timestamp(t)
+    }
 }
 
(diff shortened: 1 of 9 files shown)


src/journal/src/stream.rs
```diff
@@ -14,37 +14,38 @@
...
+use serde::{Deserialize, Serialize};
 
 use super::async_trait;
 
-/// A generic timestamp to order events.
-pub trait Timestamp:
-    Ord + Send + Sync + Copy + Debug + Unpin + Serialize + DeserializeOwned
-{
-}
+#[derive(PartialEq, PartialOrd, Clone, Copy, Debug, Serialize, Deserialize)]
```
tisonkun, line 21: I'd still derive `Serialize` and `Deserialize` for it to avoid assuming a Timestamp is always a `u64`.

huachaohuang:
@tisonkun Thanks, I am thinking about a similar change. `u64` is the most reasonable choice for now. And if that turns out to be not sufficient, we can simply use a bytewise comparable `Vec<u8>`. It is not hard to convert a `u64` to `Vec<u8>`, for example, `u64.to_be_bytes()`.

huachaohuang:
I will review this PR when I settle down things in https://github.com/engula/engula/pull/138.

tisonkun:
@huachaohuang any updates?

huachaohuang:
@tisonkun Oh, let me take a look, seems I missed your messages in #153

huachaohuang:
@tisonkun I prefer to close this in favor of #153 because even if we merge this first, we still need to fix things in #153  and then fix the grpc implementation. And I think the later fixes for the grpc implementation will be different from this PR too.

tisonkun:
@huachaohuang OK. Either way works for me.

PR #140, opened 2021-11-30
engine: add a memory hash engine

huachaohuang:
Enough for abstractions, let's build our first engine now.
This PR only handles some memory data. We will offer the hash engine an `Env` later to persist its state.
I also remove the API part to avoid confusion for now.


src/engine/hash/src/memtable.rs
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
+    pub async fn get<K: AsRef<[u8]>>(&self, key: K) -> Option<Vec<u8>> {
+        let map = self.map.lock().await;
+        map.get(key.as_ref()).cloned()
+    }
+
+    pub async fn set<K: Into<Vec<u8>>, V: Into<Vec<u8>>>(&self, key: K, value: V) {
+        let key = key.into();
```
zojw, line 41: is it a internalKey(with seqNo or some mvcc version) or user key?

if it's user key, maybe we need maintain self.size like `size = size + new_value - old_value` when set exist key-value pair twice :)

huachaohuang: Oh, I haven't considered MVCC at this moment. So user keys and internal keys are the same things for now.

src/engine/hash/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        let got = engine.get(&key).await?;
+        assert_eq!(got, Some(value));
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn table() -> Result<()> {
+        let records = vec![(vec![1], vec![1]), (vec![2], vec![2])];
+
+        let filename = "/tmp/hash";
```
zojw, line 47: maybe better use `env::temp_dir()` for windows developer~

zojw (approved):
LGTM for first version and we can improve it later :)

PR #142, opened 2021-11-30
kernel: an alternative implementation

huachaohuang:
This is a new implementation of Kernel, which leverages trait bounds.

I am going to use this in https://github.com/engula/engula/pull/140.

Ref: https://github.com/engula/engula/pull/136


huachaohuang:
Close this in favor of #155

PR #143, opened 2021-12-01
engine: persist data to kernel

### LINKED ISSUE #59: Settle down the project layout

huachaohuang:
This PR enables the hash engine to persist data in a kernel. It demonstrates the interaction between `Engine` and `Kernel`.

Note that our kernel is still buggy as described [here](https://github.com/engula/engula/pull/155#discussion_r762415322), so the engine recovery thing doesn't work if we test with more records [here](https://github.com/engula/engula/pull/143/files#diff-5ba477865d081fa33459fa0e3f77c3b96da6ca5a62ae7548ce45fc2b6acc71f8R38). I will fix that on the kernel side later.

Closes #59


src/kernel/src/mem/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+use tokio_stream::wrappers::BroadcastStream;
+
+use crate::{
+    async_trait, Error, KernelUpdate, Result, ResultStream, Sequence, Version, VersionUpdate,
+};
+
+#[derive(Clone)]
+pub struct Kernel {
+    stream: MemStream,
+    bucket: MemBucket,
```
tisonkun, line 30: If the `Kernel` holds `Stream` & `Bucket` directly, what's the meaning of `Journal` & `Storage`?

huachaohuang: Oh, this is only a simplified version of `Kernel`. We still want to provide a kernel that allows users to create multiple streams and buckets. Then we will need to hold `Journal` and `Storage`.

tisonkun: Get it. I think I will go back to the code after you squash all these PRs and see what we get at that time. Comments on intermediate changes have little meaning :P

huachaohuang: Yes, just put this here for the big picture. I will submit smaller changes in future PRs to land it.

tisonkun: You can ping me as a voluntary reviewer then :)

zojw:
it seems we have multiple usage modes, the two most obvious ways are "as-storage-lib" or "as-storage-service".....

it lets me think a question:

- do we also need to handle logic like rocksdb's "OPTION-XXXX" file?
- does components assembly logic is also like "using mem or file" also need persist as config file?
- maybe info like OPTION also need in "as-storage-service mode"

huachaohuang:
> it seems we have multiple usage modes, the two most obvious ways are "as-storage-lib" or "as-storage-service".....
> 
> it let us think a question:
> 
> * do we also need to handle logic like rocksdb's "OPTION-XXXX" file?
> * does components assembly logic is also like "using mem or file" also need persist as config file?
> * maybe info like OPTION also need in "as-storage-service mode" and where are they stored?

Yes, we need to persist the kernel options once a kernel is created. It is very dangerous to use a kernel in one way and then switch to another way. So we should provide some validation when a kernel is open.

src/engine/hash/src/engine.rs
```diff
@@ -12,61 +12,241 @@
...
-        if let Some(value) = inner.mem.get(&key).await {
+    pub async fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>> {
+        let current = self.current_version().await;
+        current.get(key).await
+    }
+
+    pub async fn set(&self, ts: Timestamp, key: Vec<u8>, value: Vec<u8>) -> Result<()> {
+        let current = self.current_version().await;
+        current.set(ts, key, value).await?;
+        if let Some((imm, version)) = current.should_flush().await {
```
zojw, line 110: ps: switch memtable maybe need lock carefully~

zojw, line 110: maybe here has some concurrency problem? two thread set at the same time maybe generate two immutable table & verison...

huachaohuang: Well, that's why I acquire a lock at https://github.com/engula/engula/pull/143/files#diff-4ee1308253a5d8885bc1d6a84ecf80e15233c086174740a737302e56654ac6a2R58.

huachaohuang: We still need to prevent flushing multiple memory tables at the same time, though. There are still a lot of work to do before this engine can work decently 🤣

huachaohuang:
@zojw @tisonkun This PR is ready for review :)

tisonkun:
@huachaohuang thanks for your reminder! I'm reviewing this PR now.

src/engine/hash/src/engine.rs
```diff
@@ -12,61 +12,259 @@
...
+        for object in &version.objects {
+            let reader = bucket.new_sequential_reader(object).await?;
+            let table_reader = TableReader::new(reader).await?;
+            tables.push(Arc::new(table_reader));
+        }
+        let last_timestamp = if let Some(ts) = decode_last_timestamp(&version.meta)? {
+            ts
+        } else {
+            0
+        };
```
tisonkun, line 187:
```suggestion
        let last_timestamp = decode_last_timestamp(&version.meta)?.unwrap_or(0);
```

src/engine/hash/src/format.rs
```diff
@@ -0,0 +1,62 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use bytes::{Buf, BufMut};
+use tokio::io::{AsyncRead, AsyncReadExt};
+
+use crate::{Error, Result};
+
+pub type Timestamp = u64;
```
tisonkun, line 20: Why not use `engula_journal::Timestamp`?

src/engine/hash/src/format.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright 2021 The Engula Authors.
```
tisonkun, line 1: Can we rename this file as `data_format.rs` or `codec.rs`? `format.rs` can be interpreted as display format.

src/engine/hash/src/memtable.rs
```diff
@@ -12,41 +12,56 @@
...
-        map.get(key.as_ref()).cloned()
+    pub async fn get(&self, key: &[u8]) -> Option<Vec<u8>> {
+        let inner = self.inner.lock().await;
+        inner.map.get(key).cloned()
+    }
+
+    pub async fn set(&self, ts: Timestamp, key: Vec<u8>, value: Vec<u8>) {
+        let mut inner = self.inner.lock().await;
+        inner.size += format::record_size(&key, &value);
+        inner.last_ts = ts;
```
tisonkun, line 51: `assert!` the incoming ts greater than last_ts?

tisonkun (commented):
Thanks for preparing this PR! Comments inline.

I'll take a closer look at `engine.rs` later.

src/engine/hash/src/engine.rs
```diff
@@ -12,61 +12,259 @@
...
+        let mut version = self.clone();
+        version.last_sequence = update.sequence;
+        if let Some(ts) = decode_last_timestamp(&update.set_meta)? {
+            version.last_timestamp = ts;
+        }
+
+        for object in &update.add_objects {
+            // We assume that objects are flushed from the oldest immtable to the newest.
+            let reader = version.bucket.new_sequential_reader(object).await?;
+            let table_reader = TableReader::new(reader).await?;
```
zojw, line 247: mark a improvement~ maybe we can build all readers first then consume readers :)

logic like

```
for {
	// build readers
}
for {
	// build table reader
}
```

then for gRPC impl, it can pipeline send requests to remote

src/engine/hash/src/engine.rs
```diff
@@ -12,61 +12,259 @@
...
+        Ok(engine)
     }
-}
 
-impl Engine {
-    pub fn new() -> Self {
-        let inner = Inner {
-            mem: Arc::new(Memtable::new()),
-            imm: Vec::new(),
+    async fn recover(&self) -> Result<()> {
```
zojw, line 78: maybe `last_number` also need be recover in recover() or open() method?

src/engine/hash/src/format.rs
huachaohuang: `engula_journal::Timestamp` is a type that I can't manipulate (e.g. add/sub). It is possible to make that work, though. But I am not clear about the relationship between engine timestamp and journal timestamp yet. For example, they may not always be the same thing. So I introduce another timestamp in the engine for now.

src/engine/hash/src/format.rs
huachaohuang: Well, I prefer the short one `codec.rs` then.

src/engine/hash/src/engine.rs
huachaohuang: Sounds a good idea.

src/engine/hash/src/engine.rs
huachaohuang: Oh, I think you're right.

src/engine/hash/src/engine.rs
```diff
@@ -12,61 +12,263 @@
...
+        let current = self.current_version().await;
+        let mut ts = current.last_timestamp;
+        let mut stream = self.stream.read_events((ts + 1).into()).await;
+        while let Some(events) = stream.try_next().await? {
+            for event in events {
+                ts += 1;
+                let (key, value) = codec::decode_record(&event.data)?;
+                current.put(ts, key.to_owned(), value.to_owned()).await;
```
zojw, line 83: do we need to check MEMTABLE_SIZE here after recovery~?

it seems some [rocksdb-like engines](https://github.com/cockroachdb/pebble/blob/cbcf6e0d2f38d18a82aa50e48bdb28c192b90461/open.go#L591) do that(it also has smaller memtable size threshold than normal flush operation, restart process maybe produce more sst file - -), I'm not sure is any practice need do it 😕

src/engine/hash/src/format.rs
tisonkun: OK. You can resolve these comments.

src/engine/hash/src/engine.rs
```diff
@@ -40,8 +37,8 @@ pub struct Engine<K: Kernel> {
     stream: K::Stream,
     bucket: K::Bucket,
     current: Arc<Mutex<Arc<EngineVersion<K>>>>,
-    last_ts: Arc<Mutex<Timestamp>>,
-    last_number: Arc<AtomicU64>,
+    last_timestamp: Arc<Mutex<Timestamp>>,
```
tisonkun, line 40: why do you rename the field to `last_timestamp`?

src/engine/hash/src/format.rs
tisonkun: After a closer look, I agree that we hold a different type here. Engine is a user of Journal, and the reason we create a newtype for Journal's timestamp is for allowing users using their own timestamp and when talking to the Journal, adapt it to Journal's timestamp. These lines are just the case.

src/engine/hash/src/engine.rs
huachaohuang: Hmm, I can't give a strong reason here. Just thought that if I have multiple `last_xxx` fields here, maybe I'd better make them more verbose.

src/engine/hash/src/engine.rs
huachaohuang: Yes, I think so. But maybe leave it as a future improvement. This PR has done enough.

zojw (approved):
LGTM

src/engine/hash/src/engine.rs
huachaohuang: @tisonkun Any more comments? If not, I am going to merge it then.

tisonkun (approved):
LGTM.

huachaohuang:
Thanks for your review. I believe this is a big step to complement the bridge between `Engine` and `Kernel` :)

src/engine/hash/src/engine.rs
```diff
@@ -12,61 +12,263 @@
...
+
+        let mut table_builder = TableBuilder::new(writer);
+        for (key, value) in imm.iter().await {
+            table_builder.add(&key, &value).await?;
+        }
+        table_builder.finish().await?;
+
+        let mut update = KernelUpdate::default();
+        let last_ts = encode_u64_meta(imm.last_update_timestamp().await);
+        update.set_meta(LAST_TIMESTAMP, last_ts);
```
zojw, line 133: after some learning, it seems we'd better add check `last_ts > [current_last_ts]` before do update.

someone ingest SST maybe contains higher ts value than `imm.last_update_timestamp`

but it's ok for now, because we don't support ingest 😄

huachaohuang: Yeah, the engine is quite buggy for now.

PR #154, opened 2021-12-01
storage: cleanup traits for v0.2

huachaohuang:
Check #143 for example usage of Storage.


src/storage/src/mem/bucket.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+pub struct SequentialReader {
+    object: Object,
+    offset: usize,
+}
+
+impl SequentialReader {
+    fn new(object: Object) -> Self {
+        Self { object, offset: 0 }
```
zojw, line 81: maybe we still need export offset as parameter in furture? (just a question not review

zojw (commented):
LGTM, I'd like to fix file impl tomorrow

zojw (approved):
LGTM, I'd like to fix file impl tomorrow

src/storage/src/mem/bucket.rs
huachaohuang: Yes, maybe. Or a random reader.

PR #153, opened 2021-12-01
journal: cleanup traits for v0.2

huachaohuang:
Check https://github.com/engula/engula/pull/143 for example usage of Journal.

Issue #146, opened 2021-12-01
A memory kernel that stores everything in memory


huachaohuang:
A memory kernel integrates a memory journal, storage, etc.

huachaohuang:
Closed by #155

PR #155, opened 2021-12-02
kernel: add abstractions and a memory implementation

### LINKED ISSUE #149: Provide a kernel abstraction for engines
A kernel is responsible for the persistent part of an engine. It provides the ability to store events, objects, metadata, etc.

huachaohuang:
This PR is based on https://github.com/engula/engula/pull/153. Check #143 for example usage of Kernel.

Close #149 #146 

I realized that maybe we can generalize a local kernel and provide a memory kernel and file kernel as type alias. For example:

```rust
pub mod local {
    pub struct Kernel<S: Stream, B: Bucket> { ... }
}

pub mod mem {
    pub use engula_journal::mem::Stream;
    pub use engula_storage::mem::Bucket;
    pub type Kernel = super::local::Kernel<Stream, Bucket>;
}

pub mod file {
    pub use engula_journal::file::Stream;
    pub use engula_storage::file::Bucket;
    pub type Kernel = super::local::Kernel<Stream, Bucket>;
}
```

~~Maybe we can try it in next PR.~~ I have already added the idea in this PR :)


src/kernel/src/kernel.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::{async_trait, Bucket, Result, ResultStream, Sequence, Stream, Version, VersionUpdate};
+
+/// An interface to interact with a kernel.
+#[async_trait]
+pub trait Kernel: Clone + Send + Sync + 'static {
+    type Stream: Stream;
+    type Bucket: Bucket;
+
+    /// Returns a journal stream.
+    async fn stream(&self) -> Result<Self::Stream>;
```
zojw, line 26: this only return 1 stream/ 1 storage, so it's journal's duty to handle multiple stream/storage situation?

for example: machine has two disks and we want to shard write workload to two disks, so it need implement a virtual stream to combine two local streams: route write and merge read? 

maybe some double-write situtation also need virutal stream/storage to ensure both success or failure (or can delete obsolete info, e.g. write s3 and local but failed in local)?

huachaohuang: Yeah, someone asked the same question, check [here](https://github.com/engula/engula/pull/143#discussion_r759892162) :)

huachaohuang:
@zojw @tisonkun This PR is ready. And I think other tasks for v0.2 depend on the `Kernel` added here. So please help to review this when possible so that we can move forward :)

tisonkun:
@huachaohuang thanks for your reminder! I'm reviewing this now.

src/kernel/src/kernel.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// Returns a storage bucket.
+    async fn bucket(&self) -> Result<Self::Bucket>;
+
+    /// Installs a kernel update.
```
tisonkun, line 31: I'm somehow confused by the name `install_update`. Is there a reference to the origin? I saw similar concepts named `mutations`, `multiop`, etc.

src/kernel/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+mod kernel;
+mod version;
+
+mod local;
+
+pub use async_trait::async_trait;
+pub use engula_journal::{Stream, Timestamp};
+pub use engula_storage::Bucket;
+
+pub type ResultStream<T> = Box<dyn futures::Stream<Item = Result<T>> + Send + Unpin>;
```
tisonkun, line 41: Does it logically equal to `engula_jorunal::ResultStream`? If so, we should not redefine one more type for the same purpose.

src/kernel/src/local/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+
+//! A [`Kernel`] implementation that stores everything in memory.
+//!
+//! [`Kernel`]: crate::Kernel
+
+mod kernel;
+
+pub mod mem;
+
+pub use self::kernel::Kernel;
```
tisonkun, line 23: Can we add some tests to reflect basic usage?

src/kernel/src/local/kernel.rs
```diff
@@ -0,0 +1,92 @@
...
+    async fn bucket(&self) -> Result<Self::Bucket> {
+        Ok(self.bucket.clone())
+    }
+
+    async fn install_update(&self, update: KernelUpdate) -> Result<()> {
+        let mut inner = self.inner.lock().await;
+        inner.last_sequence += 1;
+        let mut version_update = update.update;
+        version_update.sequence = inner.last_sequence;
+        // TODO: update current version
```
tisonkun, line 73: There is `inner.last_sequence += 1;` above. What do you mean by "TODO: update current version"?

src/kernel/src/local/kernel.rs
```diff
@@ -0,0 +1,92 @@
...
+        inner
+            .updates
+            .send(Arc::new(version_update))
```
tisonkun, line 76: I'm not quite familiar with this part. Then it seems the current `Kernel` just receive all updates and pipe it when someone subscribes via `version_updates`. When do the inner stream and bucket apply updates?

tisonkun (commented):
Thanks for preparing this pull request. Comments inline.

src/kernel/src/kernel.rs
huachaohuang: I model `Kernel` from [RocksDB's VersionSet](https://github.com/facebook/rocksdb/blob/main/db/version_set.h), but I make `Kernel` more general to support different implementations. I think the name is my personal taste, though. You can read it literally, like "install an update to the kernel to get a new version".

src/kernel/src/lib.rs
huachaohuang: They are exactly the same. I think we may need to add a "common" or "util" module to share these things sooner or later.

src/kernel/src/local/kernel.rs
huachaohuang: Oh, you see, the `inner.current` version is never updated. We always return an empty version in `current_version()`. So this kernel implementation only work when you first open it 🤣

src/kernel/src/local/kernel.rs
huachaohuang: Ther kernel's job here is to persist the update (part of #164) and pipe it to whoever needs it. I think the best way to illustrate the usage is to check #143 as well.

src/kernel/src/kernel.rs
tisonkun: Thanks for your explanation. The definition makes sense.

However, if a concept is totally new, every user will ask the same question later as the project grows. I tend to avoid creating new concepts especially that "install" has a strong impact on "installing software".

An alternative could be:

```rust
/// Apply a series of version updates.
async fn apply(&self, update: KernelUpdate) -> Result<()>;
```

You can still name it as `install` and observe how it impact contributors to understand the underneath.

src/kernel/src/lib.rs
tisonkun: Make sense to factory out basic crates. Let me emphasize that these crates are defined as basic dependencies which have as few dependencies as possible.

The importance of such a definition is that otherwise, developers tend to throw everything into a "util" or "common" crate and the crate grows quickly and ugly, e.g., TiDB's "util" directory. And for this sake, I'd prefer "core" or "base" than "util"/"common" which highlights that fewer dependencies are preferred.

tisonkun: Some references [here](https://twitter.com/tison1096/status/1417750125196189698).

src/kernel/src/kernel.rs
huachaohuang: How about `apply_update`? A simple `apply` may be a bit vague as `Kernel` plays multiple roles.

tisonkun: That sounds good.

src/kernel/src/lib.rs
huachaohuang: Well, "util" and "common" are the most appropriate names I ever heard for this. I have no context about the reference you gave. But I guess it is about bad management, not wrong names. Bad management ruins everything no matter how you name it. But it's not the name's fault.
For us, I think the rule is simple, if two crates need the same thing, you can put it in a common crate.

src/kernel/src/local/kernel.rs
tisonkun: Thanks! So "update current version" here isn't "version-number", but "versioned data". Is there a follow-up PR for implementing this?

PS: You can resolve this comment.

src/kernel/src/local/kernel.rs
tisonkun: Thanks for pointing that out. So Engula adopts a strong client to manipulate low-level concepts, instead of a lightweight client only sending high-level requests and letting the server do all thing underneath. It can work and let's see what it brings to us :)

tisonkun (approved):
LGTM. Basic usage can be covered by #143. However, we can always add kernel tests just in the crates later :P

src/kernel/src/lib.rs
tisonkun: For `ResultStream` itself, if `ResultStream` here means the type that when you read events from a stream of journal returns, you can `pub use engula_journal::ResultStream` and prevent redefining it.

huachaohuang: Both `ResultStream`s here and in `engula-journal` are simply a common type alias that doesn't imply any logic. So I think we can live with it now and see if a common crate is necessary later.

src/kernel/src/local/kernel.rs
huachaohuang: Yes, I think it will be part of https://github.com/engula/engula/issues/164

src/kernel/src/lib.rs
huachaohuang: Oh, I am wrong with this, the `ResultStream` actually contains a crate-specific `Error`, so it is something like the `Result` in every crate.

src/kernel/src/local/kernel.rs
huachaohuang: Hmm, I think you may misunderstand something here, but I can't tell 🤔

src/kernel/src/lib.rs
tisonkun: OK.

tisonkun (approved):

huachaohuang:
OK, let's merge this and call it a day :)

PR #158, opened 2021-12-02
journal: adapt grpc journal into new abstraction 

tisonkun:
This is a follow up to #153 .


tisonkun:
@huachaohuang gRPC journal fixed.

src/journal/src/grpc/journal.rs
```diff
@@ -12,38 +12,34 @@
...
-
 use super::{
     client::Client,
-    error::Result,
     proto::{CreateStreamRequest, DeleteStreamRequest},
-    stream::RemoteStream,
+    stream::Stream,
 };
-use crate::{async_trait, Journal, Timestamp};
+use crate::{async_trait, error::Result};
```
huachaohuang, line 20: I am OK with `crate::Result` now, so you may consider it :)

src/journal/src/grpc/journal.proto
```diff
@@ -65,6 +65,10 @@ message ReadEventRequest {
 }
 
 message ReadEventResponse {
```
huachaohuang, line 67: s/ReadEvent/ReadEvents, let's keep it consistent with the trait.

src/journal/src/grpc/proto.rs
```diff
@@ -14,13 +14,12 @@
 
 tonic::include_proto!("engula.journal.grpc.v1");
```
huachaohuang, line 15:
```suggestion
tonic::include_proto!("engula.journal.v1");
```

I think "engula.journal.v1" is sufficient, the "grpc" is redundant.

src/journal/src/grpc/server.rs
```diff
@@ -12,61 +12,39 @@
...
-impl<S, J, T> Server<S, J, T>
+impl<J> Server<J>
 where
-    T: Timestamp + 'static,
-    S: Stream<Timestamp = T> + Send + Sync + 'static,
-    S::Error: Send + Sync + 'static,
-    S::EventStream: Send + Sync + 'static,
-    J: Journal<S> + Send + Sync + 'static,
-    Status: From<S::Error>,
+    J: Journal + Send + Sync + 'static,
```
huachaohuang, line 30: You can simply `impl<J: Journal> Server<J>`, `Journal` already implies `Send + Sync + 'static` now.

src/journal/src/grpc/server.rs
```diff
@@ -12,61 +12,39 @@
...
-impl<S, J, T> journal_server::Journal for Server<S, J, T>
+impl<J> journal_server::Journal for Server<J>
 where
-    T: Timestamp + 'static,
-    S: Stream<Timestamp = T> + Send + Sync + 'static,
-    S::Error: Send + Sync + 'static,
-    S::EventStream: Send + Sync + 'static,
-    J: Journal<S> + Send + Sync + 'static,
-    Status: From<S::Error>,
+    J: Journal + Send + Sync + 'static,
```
huachaohuang, line 44: ditto

src/journal/src/grpc/mod.rs
```diff
@@ -13,35 +13,28 @@
...
+    use super::Server;
     use crate::*;
 
     #[tokio::test(flavor = "multi_thread")]
     async fn test() -> std::result::Result<(), Box<dyn std::error::Error>> {
         let listener = TcpListener::bind("127.0.0.1:0").await?;
         let local_addr = listener.local_addr()?;
         tokio::task::spawn(async move {
-            let j: mem::MemJournal<u64> = mem::MemJournal::default();
+            let j: mem::Journal = mem::Journal::default();
```
huachaohuang, line 37:
```suggestion
            let j = mem::Journal::default();
```

src/journal/src/grpc/proto.rs
```diff
@@ -14,13 +14,12 @@
...
 
-use super::error::Error;
-use crate::Timestamp;
+use crate::{error::Error, Timestamp};
 
-pub fn serialize_ts<T: Timestamp>(ts: &T) -> Result<Vec<u8>, Error> {
+pub fn serialize_ts(ts: &Timestamp) -> Result<Vec<u8>, Error> {
```
huachaohuang, line 19: Serializing a timestamp to json really bugs me and also makes the related code complicated. Let me think about it.

huachaohuang: Maybe we consider it later. We can keep it this way for now to avoid mixing things in the same PR.

tisonkun: Yep. Here is an abstraction saying we should serialize the timestamp. How to do it effectively can be considered. I'm unfamiliar with serde so it takes some time to get used to it and find an optimization.

huachaohuang:
BTW, you can add the grpc implementation [here](https://github.com/engula/engula/blob/main/src/journal/src/lib.rs#L27).

src/journal/src/error.rs
```diff
@@ -23,6 +23,26 @@ pub enum Error {
     AlreadyExists(String),
     #[error("{0}")]
     InvalidArgument(String),
+    #[error(transparent)]
+    GrpcStatus(#[from] tonic::Status),
```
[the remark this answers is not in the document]
tisonkun: Yes. I'm thinking of it. But what general errors should we name it? A fallback is `Internal`/`Unknown`, but I don't get a good idea.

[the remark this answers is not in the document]
huachaohuang: I think we can simply convert everything we haven't handled explicitly to `Unknown` for now.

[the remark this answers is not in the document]
tisonkun: Updated.

src/journal/src/grpc/journal.rs
tisonkun: Updated.

src/journal/src/grpc/journal.proto
tisonkun: Updated.

src/journal/src/grpc/proto.rs
tisonkun: OK.

tisonkun:
@huachaohuang comments addressed.

huachaohuang (commented):
Thanks, it looks much better now. One more thing, can you add a simple description like [this](https://github.com/engula/engula/blob/main/src/journal/src/lib.rs#L26) and [this](https://github.com/engula/engula/blob/main/src/journal/src/mem/mod.rs#L15)?

tisonkun:
@huachaohuang updated.

huachaohuang (approved):

Issue #150, opened 2021-12-01
Add a simple command to do local checks before sending PRs


huachaohuang:
We need to run several commands to check tests, style, license, unused dependencies before sending a PR, which is tedious. Maybe we can add a simple way to check everything at once locally.

huachaohuang:
A Makefile is a possible solution, but maybe there is some Rusty solution too.

tisonkun:
I don't meet this requirement and if you need it, you can start with writing your local script.

Basically, the command is documented in the CONTRIBUTING.md and the details and almost oneliners:

```
cargo build
cargo test --workspace
cargo clippy --workspace --tests --all-features -- -D warnings
cargo fmt --all -- --check
```

while several workflows require pre-install binary:

```
cargo audit
cargo udeps--workspace
# apache/skywalking-eyes
# editorconfig-checker/editorconfig-checker
```

tisonkun:
In most cases I meet, CLion or any proper configure IDE can help you to format code. Audit, udeps, and even licenses violation are rare.

tisonkun:
Also, there is a utility [`nektos/act`](https://github.com/nektos/act) to run GitHub Actions locally, although it will setup a container environment which cost some time, but it's the exact workflow running by GitHub Actions.

tisonkun:
If we really need one, either a make target or cargo customized target (if possible, cc @PsiACE) is OK. But to prevent the command out-of-date, we must update the CI workflows to reflect the exact command logics. Otherwise we'll quickly run into an out of sync situation.

PsiACE:
If it's a mandatory check, maybe try pre-commit or adding githooks. If it is optional, consider using a makefile or a custom cargo command.

> But to prevent the command out-of-date, we must update the CI workflows to reflect the exact command logics.

:+1:

huachaohuang:
How does a custom cargo command work in our use cases?

PsiACE:
> How does a custom cargo command work in our use cases?

add a cargo cmd alias like this

```
[alias]
test-all = "test --workspace"
clippy-all = "clippy --workspace --tests --all-features -- -D warnings"
fmt-check = "fmt --all -- --check"
```

huachaohuang:
@PsiACE Can it combine multiple commands into one? Otherwise, it doesn't seem very helpful, I still need to run multiple commands separately.

PsiACE:
[...]

PR #159, opened 2021-12-02
makefile: intro makefile to simplify the local workflow

PsiACE:
Fixes: #150

add a basic makefile to simplify the local workflow

- make setup: setup all crates
- make lint: fmt check and clippy
- make audit: security audit and find unused deps
- make pre-commit: run all local workflow like ci

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+setup:
+	cargo install cargo-audit cargo-udeps
+
+audit:
+	cargo audit
+	cargo udeps --workspace
+
+clean:
+	cargo clean
+
+build:
+	cargo build
+
+clippy:
+	cargo clippy --workspace --tests --all-features -- -D warnings
+
+fmt:
+	cargo fmt
+
+fmt-check:
+	cargo fmt --all -- --check
+
+lint: fmt-check clippy
+
+pre-commit: lint build audit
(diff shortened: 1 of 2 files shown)


tisonkun (commented):
If we'd like to add a pre-commit script, then let's directly do it, instead of adding a `Makefile` and assume how users will use it.

Also, it doesn't cover editor config checking and license auditing. I don't like to pull something random into the codebase before figuring out we really want it. If some developers need a handy script to verify a patch before submitting it, I suggest they write their own script and share it on the discussion forum. The difference here is that we don't maintain this script on behalf of the project, but leave it to developers unofficially.

PsiACE:
> If we'd like to add a pre-commit script, then let's directly do it, instead of adding a Makefile and assume how users will use it.
This makefile also simplifies other commands.
> Also, it doesn't cover editor config checking and license auditing
We can add these checks. Databend also uses a makefile cmd to help format the shell and yaml.
> I don't like to pull something random into the codebase before figuring out we really want it. 
Nothing is really necessary. Just some sugar. In fact, experienced developers don't rely on this. But anyone would like to be able to take it easy.

PsiACE:
@tisonkun convinced me that we should avoid introducing some unnecessary complexity. Unless we are faced with some tricky/complicated situation, there is no need to do anything redundant.

huachaohuang:
@PsiACE @tisonkun Thanks for both of your work. I don't have a strong opinion about this, so I think we can keep it as it is until we have better ideas :)

PR #156, opened 2021-12-02
storage: fix file storage after clean up v0.2 trait

zojw:
simple fix file storage impl after #154

PR #165, opened 2021-12-04
storage: fix gRPC storage after clean up v0.2 trait

### LINKED ISSUE #68: Roadmap v0.2 - Storage

zojw:
follow up to https://github.com/engula/engula/pull/156#issuecomment-984431569, closes #68 

fix gRPC storage after clean up v0.2 trait


tisonkun:
@zojw you can update branch since #155 merged.

huachaohuang:
I think this closes #68 .

tisonkun:
@zojw thanks for your updating. I'm reviewing this now. Next time you can ping me after updates.

huachaohuang:
I submit a small fix https://github.com/zojw/engula/pull/1

src/storage/src/grpc/client.rs
```diff
@@ -38,11 +39,25 @@ impl Client {
 
     method!(delete_bucket, DeleteBucketRequest, DeleteBucketResponse);
```
tisonkun, line 40: Since we start customizing methods, I'd prefer to expand these macro manually.

zojw:
> I submit a small fix [zojw#1](https://github.com/zojw/engula/pull/1)

oh...thanks :)

zojw:
@huachaohuang just as your suggestion it at dinner, I try StreamReader, it seems we can simplify reader by zojw/engula#2

src/storage/src/grpc/mod.rs
```diff
@@ -42,25 +42,52 @@ mod tests {
...
                 .unwrap();
         });
+        let url = format!("http://{}", local_addr);
+        let storage = RemoteStorage::connect(&url).await?;
+        storage.create_bucket("bucket").await?;
+        let b = storage.bucket("bucket").await?;
+        let mut w = b.new_sequential_writer("object").await?;
+        let buf = vec![0, 1, 2];
+        w.write(&buf).await?;
+        w.flush().await?;
```
zojw, line 59: it has an incompatible semantic with the classic local file system...

current, users cannot get real error(not channel full) info after a flush, they only can observe real error when shutdown file - -

huachaohuang:
Added an alternative implementation for `SequentialWriter` https://github.com/zojw/engula/pull/3

huachaohuang:
> @huachaohuang just as your suggestion it at dinner, I try StreamReader, it seems we can simplify reader by [zojw#2](https://github.com/zojw/engula/pull/2)

Looks nice. I think you can try `ReaderStream` for https://github.com/engula/engula/pull/165/files#diff-6ea3f665d464f10f55fea3d6b1f803768d7651c97da59bb62f9aaf5f80fbaae7R93 too.

huachaohuang:
We can optimize the stream RPC requests later. For example, for a `ReadObject` stream, we can send an `OpenObjectRequest` first to set up the stream metadata (like bucket/object name), and then send `ReadObjectRequest`s with only object length or something.

huachaohuang:
It also applies to the `UploadObject` stream.

src/storage/src/grpc/mod.rs
huachaohuang: Yes, I think the semantic of `AsyncWrite` is not ideal. But let's live with it for this release.

src/storage/src/error.rs
```diff
@@ -25,6 +25,8 @@ pub enum Error {
     InvalidArgument(String),
     #[error(transparent)]
     Io(#[from] std::io::Error),
+    #[error("unknown: {0}")]
+    Unknown(String),
```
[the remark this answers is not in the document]
huachaohuang: I'm not sure either. So I simply accept both for now.

zojw:
> Added an alternative implementation for `SequentialWriter` [zojw#3](https://github.com/zojw/engula/pull/3)

thanks~ LGTM and learn a lot 😄 

- mpsc::channel with Req instead of bytes let us can use ReceiverStream to simplify rx -> steaming request convert
- tokio::spawn a task to consume rx & request gRPC method simplify much logic about when poll future or how to handle channel full question
- without PollSender, the previous commit seems to have the correctness problem - -

thanks, let me try ReaderStream

src/storage/src/error.rs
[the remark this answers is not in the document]
zojw: I vote to keep nested structure errors as possible 😄 , nested errors make users have their chance to control whether or how to flatten error info

for example: they can choose to omit verbose error info after servlet after many occurrences like https://yoshihisaonoue.wordpress.com/2021/02/07/jvm-option-xx-omitstacktraceinfastthrow/

src/storage/src/lib.rs
```diff
@@ -23,16 +23,20 @@
 //!
 //! Some built-in implementations of [`Storage`]:
 //!
+//! - [`file`](crate::file)
 //! - [`mem`](crate::mem)
+//! - [`grpc`](crate::grpc)
```
huachaohuang, line 28: Maybe move this up to align with the `mod` order.

src/storage/src/grpc/mod.rs
```diff
@@ -12,26 +12,26 @@
...
 
-pub use self::{
-    client::Client,
-    error::{Error, Result},
-    object::RemoteObject,
-    server::Server,
-    storage::RemoteStorage,
-    uploader::RemoteObjectUploader,
-};
+pub use self::{client::Client, server::Server, storage::Storage as RemoteStorage};
```
huachaohuang, line 26: Why rename the `Storage`?

src/storage/src/grpc/bucket.rs
```diff
@@ -0,0 +1,159 @@
...
+        };
+        self.client.delete_object(input).await?;
+        Ok(())
+    }
+
+    async fn new_sequential_reader(&self, name: &str) -> Result<Self::SequentialReader> {
+        let input = ReadObjectRequest {
+            bucket: self.bucket_name.to_owned(),
+            object: name.to_owned(),
+            // both unused for sequential reader.
```
huachaohuang, line 71: Are these two fields used in other places? If not, we can simply remove it now.

src/storage/src/error.rs
```diff
@@ -25,6 +25,8 @@ pub enum Error {
     InvalidArgument(String),
     #[error(transparent)]
     Io(#[from] std::io::Error),
+    #[error("unknown: {0}")]
```
huachaohuang, line 28: How about `#[error(transparent)]`?

huachaohuang (commented):
Rest LGTM. I think we learned a lot about streams in `Rust` :)

huachaohuang (approved):

PR #166, opened 2021-12-05
ci: add action to check TOML format

### LINKED ISSUE #120: Find tool and add Github Action to check toml files

DCjanus:
close #120


.taplo.toml
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021 The Engula Authors.
```
tisonkun, line 1: Must we have such a placeholder file? Besides, must the file be located at the root path?

tisonkun (commented):
Thanks for your contribution @DCjanus! Comments inline.

Also, I'd like to learn what rules taplo supports and what we earn by introducing such checkers. The diff doesn't show non-trivial improvements on formatting from my perspective.

.taplo.toml
DCjanus: Empty rule file is not required, and it's ok to remove this, the reason that I put this file here, is to let people know that we are using taplo without reading GitHub Action files.

DCjanus:
> The diff doesn't show non-trivial improvements on formatting from my perspective.

I'm not sure what is `improvements` for TOML format checker.

The reason that I would like to introduce this checker, as [this comment](https://github.com/engula/engula/issues/120#issue-1058270796), taplo would check the whitespace between `{}`, and that keep things consistent for all contributors.

.taplo.toml
tisonkun: I think you can try to figure out which [options](https://taplo.tamasfe.dev/configuration/#formatting-options) we can turn on so that it's explicit what we gain from introducing such a CI task.

DCjanus: Actual, I think the purpose to introducing a format check is about `consistent ` rather than rules, and default options are ok to me.

tisonkun (approved):
Verified locally. LGTM and it resolves the original problem about `{ xxx` and `{xxx`.

I think several improvements can be made at the upstream taplo such as customize config file location and full failure report (so far, it reports "there are some errors" and you don't know what they are but run `taplo format` to fix). But they're not blockers for this PR.

cc @huachaohuang

.taplo.toml
tisonkun: For `improvements`, for example, with default settings, we resolve the original problem about `{ xxx` and `{xxx` (BTW which rule takes effect here?).

huachaohuang: Let's remove this file if it is not required.

huachaohuang:
Wow, I found that my TOML extension for VSCode is backed by Taplo :)

huachaohuang:
BTW, I am using [Even Better TOML](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml).

.taplo.toml
huachaohuang: @DCjanus Any update? We are almost there.

DCjanus: > @DCjanus Any update? We are almost there.

Sorry for missing last message for two days. I've just deleted the empty config file

DCjanus:
And, maybe we should run `cargo install --debug taplo-cli` to reduce CI time, and carbon emissions :)

.taplo.toml
huachaohuang: Don't worry, I can configure my editor with 2 whitespaces instead.

huachaohuang:
@DCjanus Thanks for the fixes, merge now.

.taplo.toml
DCjanus: > Don't worry, I can configure my editor with 2 whitespaces instead.

To be honest, I like 4 whitespaces

huachaohuang: Oops, I do too. But maybe keep it default for peace.

huachaohuang:
Oh, I found a Cargo.toml style guide [here](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/cargo.md). TL;DR, we are not following the guide :(

Issue #148, opened 2021-12-01
A grpc kernel that stores everything in a remote grpc kernel service


huachaohuang:
A grpc kernel integrates a grpc journal, storage, etc. It consists of a client and a server part. We need to provide a binary to start the kernel server.

tisonkun:
Will the binary be part of `engula` binary? I'm thinking of deliver an one-for-all binary as [cockroachdb](https://github.com/cockroachdb/cockroach/tree/master/pkg/cli) does.

huachaohuang:
> Will the binary be part of `engula` binary? I'm thinking of delivering a one-for-all binary as [cockroachdb](https://github.com/cockroachdb/cockroach/tree/master/pkg/cli) does.

Yes, of course. I hate to operate a bunch of different binaries too.

huachaohuang:
This can be started after #167

PatrickNicholas:
I am interested in this job, I need some time to learn the details first, and the give my design.

PatrickNicholas:
I am interested in this job, I need some time to learn the details first, and the give my design.

huachaohuang:
@PatrickNicholas Thanks for your interest. It sounds good to me. The job may not be easy, so let me make it more clear:

- The grpc kernel service should accept another kernel implementation as its backend. For example, we can pass in a memory kernel to get a kernel service that stores everything in memory.
- We should provide a binary (`engula`) to start the kernel service.

This can be done in one or two PRs.

Note that this job is included in v0.2, and we aim to release that on Dec 17. So please let me know if you meet any problems. And I may take over if there is no progress until Dec 11.

PatrickNicholas:
@huachaohuang Thanks.

PatrickNicholas:
It is sad that this issue was automatically closed, due to the closing keywords set in previous PR. Shall I reopen it?

tisonkun:
@PatrickNicholas aha. I think the next time you can use `ref #xxx` where `ref` is not a keyword so that GitHub only show the reference but not close the issue automatically. I'm unaware that you separate the kernel implementation part and the cli part.

huachaohuang:
I think it's more meaningful to allow users to start independent kernel/journal/storage servers and connect them together even for v0.2. The kernel/journal/storage services are already landed, so I think we can deliver such kind of flexibility in v0.2.

For example, we can support the following commands:

```
engula journal start 127.0.0.1:10001 [--mem | --file path]
[...]

Issue #147, opened 2021-12-01
A file kernel that stores everything in local files


huachaohuang:
A file kernel integrates a file journal, storage, manifest, etc.

huachaohuang:
This can be started after #167

ZhongliGao:
Please assign this task to me and I will give the design as soon as possible

tisonkun:
> Please assign this task to me and I will give the design as soon as possible

@ZhongliGao if so, I think you can just post the ideas without assigned. Being assigned is no more than a hint and your comment here already takes effect.

tisonkun:
@huachaohuang is this issue also blocked by #128?

huachaohuang:
> @huachaohuang is this issue also blocked by #128?

Oh, I think so. But it's ok to prepare for it. The kernel part is not hard, it will be similar to a type alias like the memory kernel. So I think most work here is about the manifest, which needs to persist the metadata in local files.

PR #167, opened 2021-12-06
kernel: add manifest abstraction and a memory implementation

### LINKED ISSUE #164: Manifest abstraction to store kernel metadata
`Kernel` needs a metadata store for its metadata persistence. We have two choices:

- Add a general metadata abstraction in the same level with `Journal` and `Storage`
- Add a sub-module inside `Kernel` that only handles kernel specific metadata

For v0.2, we can choose the second solution for simplicity. We can still move to the first solution in the future if the second one is proofed to be inadequate.

huachaohuang:
Closes #164 

This PR uses protobuf to generate metadata so that we can serialize/deserialize it for storage and transport.


huachaohuang:
@tisonkun @zojw This PR is ready. It is the prerequisite of #147 and #148 .

src/kernel/src/local/kernel.rs
```diff
@@ -67,10 +78,14 @@ impl<S: Stream, B: Bucket> crate::Kernel for Kernel<S, B> {
 
     async fn apply_update(&self, update: KernelUpdate) -> Result<()> {
         let mut inner = self.inner.lock().await;
-        inner.last_sequence += 1;
+
+        let mut version = (*inner.current).clone();
```
zojw, line 82: it reads inner.current in this tasks, but install new  inner.current in other task `handle_version_updates`?

do we have chance that "first flush's version hasn't installed but new flush coming?"

zojw: emmm..maybe it has recorded by https://github.com/engula/engula/pull/143#discussion_r760111411

zojw (approved):
LGTM

src/kernel/src/local/kernel.rs
zojw: ignore me... I am wrong 🤣

tisonkun (approved):
LGTM. Although I'm a bit unsure whether we should create a layout:

```sh
local # abstract
mem # impl
```

at the same level. It's not a blocker for merging this PR.

huachaohuang:
> LGTM. Although I'm a bit unsure whether we should create a layout:
> 
> ```shell
> local # abstract
> mem # impl
> ```

Yeah, the truth is, I am not sure about that either.

PR #170, opened 2021-12-07
build: add dockerfile to build image for amd64/arm64

zojw:
simple add two docker files to build the image to help develop basic operator

then we can manually build image by 

```
DOCKER_BUILDKIT=1 docker build -f amd64.dockerfile -t engula:latest .
```

or 

```
DOCKER_BUILDKIT=1 docker build -f arm64.dockerfile -t engula:latest .
```

it will take `/bin/engula.rs` as entry method, then we can pass different subcommand to start different service in future.

at last, we can setup github workflow to release docker image to https://hub.docker.com/u/engula when code stable


PsiACE:
emmm, before introducing docker, you can take a look at https://github.com/rust-embedded/cross

huachaohuang:
Thanks, I am not familiar with these yet. So I need some time to see what we need here.

tools/ci/licenserc.yml
```diff
@@ -26,3 +26,6 @@ header:
     - 'rust-toolchain.toml'
     - 'rustfmt.toml'
     - '.cargo/audit.toml'
+    - 'amd64.dockerfile'
+    - 'arm64.dockerfile'
```
tisonkun, line 30: I think these two files deserve a license header. See also [this example](https://github.com/apache/skywalking-eyes/blob/30367d8286e324d5efc58de4c70c37ea3648306d/Dockerfile).

tisonkun (commented):
> simple add two docker files to build the image to help develop basic operator

I'd prefer introducing such docker files if we have a viable approach to develop a demo operator. Otherwise it can stay in you own fork and if you think these files help, share them in the discussion forum.

zojw:
> > simple add two docker files to build the image to help develop basic operator
> 
> I'd prefer introducing such docker files if we have a viable approach to develop a demo operator. Otherwise it can stay in you own fork and if you think these files help, share them in the discussion forum.

ok, let me attach it to operator PR

zojw:
but I think it also helps to avoid duplicate work if anyone want to run the process in docker without managed operator like kernel service in #177 (

tisonkun:
@zojw you can create an issue for such an actionable task and elaborate the motivation a bit. Barely pushing two files without background/doc is less likely to help others.

tisonkun:
Since the current binary can only print hello, I can imgaine anyone wants a docker image to do valuable things. If there is, you can share your case in an issue first.

zojw:
> Since the current binary can only print hello, I can not imagine anyone wants a docker image to do valuable things. If there is, you can share your case in an issue first.

I think "two people can avoid doing the same thing is valuable", and "current print  hello" doesn't mean giving a more way to  exec is useless after #177, it's not "print hello" anymore

but it's surely not valuable to fire an issue, so it's not valuable to give PR, so go ahead and ignore this plz :)

PR #171, opened 2021-12-07
kernel: add gRPC kernel client and server

### LINKED ISSUE #148: A grpc kernel that stores everything in a remote grpc kernel service

PatrickNicholas:
Resolves #148

### DIFF
--- a/src/kernel/build.rs
+++ b/src/kernel/build.rs
@@ -14,5 +14,7 @@
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     tonic_build::compile_protos("src/metadata.proto")?;
+    tonic_build::configure().compile(&["src/grpc/kernel.proto"], &["src"])?;
+
     Ok(())
 }
--- a/src/kernel/src/grpc/kernel.proto
+++ b/src/kernel/src/grpc/kernel.proto
@@ -1,0 +1,47 @@
+// Copyright 2021 The Engula Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package engula.kernel.v1;
+
+import "metadata.proto";
+
+service Kernel {
+    rpc ApplyUpdate(ApplyUpdateRequest) returns (ApplyUpdateResponse) {}
+
+    rpc CurrentVersion(CurrentVersionRequest) returns (CurrentVersionResponse) {}
+
+    rpc VersionUpdates(VersionUpdatesRequest) returns (stream VersionUpdatesResponse) {}
+}
+
+message ApplyUpdateRequest {
+    metadata.v1.VersionUpdate version_update = 1;
+}
+
+message ApplyUpdateResponse {}
+
+message CurrentVersionRequest {}
+
+message CurrentVersionResponse {
+    metadata.v1.Version version = 1;
+}
+
+message VersionUpdatesRequest {
+    uint64 sequence = 1;
+}
+
+message VersionUpdatesResponse {
+    metadata.v1.VersionUpdate version_update = 1;
+}
--- a/src/kernel/src/grpc/kernel.rs
+++ b/src/kernel/src/grpc/kernel.rs
@@ -1,0 +1,106 @@
+// Copyright 2021 The Engula Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+use engula_journal::{grpc as grpc_journal, Error as JournalError, Journal};
+use engula_storage::{grpc as grpc_storage, Error as StorageError, Storage};
+use futures::StreamExt;
+
+use super::{client::Client, proto::*};
+use crate::{async_trait, Error, KernelUpdate, Result, ResultStream, Sequence};
+
+pub(crate) const DEFAULT_STREAM: &str = "DEFAULT";
+pub(crate) const DEFAULT_BUCKET: &str = "DEFAULT";
+
+#[derive(Clone)]
+pub struct Kernel {
+    client: Client,
+    journal: grpc_journal::Journal,
+    storage: grpc_storage::Storage,
+}
+
+impl Kernel {
+    pub async fn connect(addr: &str) -> Result<Kernel> {
+        let client = Client::connect(addr).await?;
+        let journal = grpc_journal::Journal::connect(addr).await?;
+        let storage = grpc_storage::Storage::connect(addr).await?;
+        Ok(Kernel {
+            client,
+            journal,
+            storage,
+        })
+    }
+}
+
+#[async_trait]
+impl crate::Kernel for Kernel {
+    type Bucket = grpc_storage::Bucket;
+    type Stream = grpc_journal::Stream;
+
+    /// Returns a journal stream.
+    async fn stream(&self) -> Result<Self::Stream> {
+        match self.journal.stream(DEFAULT_STREAM).await {
+            Ok(stream) => Ok(stream),
+            Err(JournalError::NotFound(_)) => {
+                Ok(self.journal.create_stream(DEFAULT_STREAM).await?)
+            }
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    /// Returns a storage bucket.
+    async fn bucket(&self) -> Result<Self::Bucket> {
+        match self.storage.bucket(DEFAULT_BUCKET).await {
+            Ok(bucket) => Ok(bucket),
+            Err(StorageError::NotFound(_)) => {
+                Ok(self.storage.create_bucket(DEFAULT_BUCKET).await?)
+            }
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    /// Applies a kernel update.
+    async fn apply_update(&self, update: KernelUpdate) -> Result<()> {
+        let input = ApplyUpdateRequest {
+            version_update: Some(update.update),
+        };
+        self.client.apply_update(input).await?;
+        Ok(())
+    }
+
+    /// Returns the current version.
+    async fn current_version(&self) -> Result<Arc<Version>> {
+        let input = CurrentVersionRequest {};
+        let resp = self.client.current_version(input).await?;
+        let version = resp
+            .version
+            .ok_or_else(|| Error::Internal("CurrentVersionResponse::version is none".into()))?;
+        Ok(Arc::new(version))
+    }
+
+    /// Returns a stream of version updates since a given sequence (inclusive).
+    async fn version_updates(&self, sequence: Sequence) -> ResultStream<Arc<VersionUpdate>> {
+        let input = VersionUpdatesRequest { sequence };
+        match self.client.version_updates(input).await {
+            Ok(output) => Box::new(output.map(|result| match result {
+                Ok(resp) => Ok(Arc::new(resp.version_update.ok_or_else(|| {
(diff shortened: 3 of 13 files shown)


huachaohuang:
For a standalone gRPC kernel, we can run a kernel service, a journal service, and a storage service in the same gRPC server. So the client only needs one address.

huachaohuang:
Another comment, the code outside of `src/kernel/src/grpc` should not know about the gRPC implementation.

PatrickNicholas:
> For a standalone gRPC kernel, we can run a kernel service, a journal service, and a storage service in the same gRPC server. So the client only needs one address.

That means any requests to journal, storage will be send to the gRPC server of kernel? Why not support config different address for different service?

PatrickNicholas:
> Another comment, the code outside of `src/kernel/src/grpc` should not know about the gRPC implementation.

Do you mean the src/kernel/src/proto.rs should be improved by seperating the metadata and service definations? Maybe I could change the package name of metadata from `engula.kernel.v1` to `engula.meta.v1`, so the kernel service would use `engula.kernel.v1` as the package name, just like journal and storage.

huachaohuang:
> That means any requests to journal, storage will be send to the gRPC server of kernel?

Yes, in this implementation. You can regard this gRPC kernel as a standalone server that does everything itself in v0.2.

> Why not support config different address for different service?

We can do that. But maybe in the next release, I think.

huachaohuang:
> Do you mean the src/kernel/src/proto.rs should be improved by seperating the metadata and service definations? Maybe I could change the package name of metadata from `engula.kernel.v1` to `engula.meta.v1`, so the kernel service would use `engula.kernel.v1` as the package name, just like journal and storage.

Yes, that's what I mean. I think you can rename it to "engula.metadata.v1" for now. We can see if there is a better way to organize the files later.

PatrickNicholas:
@huachaohuang

I have add a journal and storage client in kernel, and use `DEFAULT-STREAM` and `DEFAULT-BUCKET` as the default value of the parameters of `stream()` and `bucket`, PTAL.

src/kernel/build.rs
```diff
@@ -12,7 +12,14 @@
...
 
+use std::path::Path;
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     tonic_build::compile_protos("src/metadata.proto")?;
+
+    let kernel_path = Path::new("src/grpc/kernel.proto");
+    let proto_dir = Path::new("src");
+    tonic_build::configure().compile(&[kernel_path], &[proto_dir])?;
+
```
huachaohuang, line 23:
```suggestion
    tonic_build::configure().compile(&["src/grpc/kernel.proto"], &["src"])?;
```

src/kernel/src/grpc/kernel.proto
```diff
@@ -0,0 +1,47 @@
...
+service Kernel {
+    rpc ApplyUpdate(ApplyUpdateRequest) returns (ApplyUpdateResponse) {}
+
+    rpc CurrentVersion(CurrentVersionRequest) returns (CurrentVersionResponse) {}
+
+    rpc VersionUpdates(VersionUpdatesRequest) returns (stream VersionUpdatesResponse) {}
+}
+
+message ApplyUpdateRequest {
+    engula.metadata.v1.VersionUpdate version_update = 1;
```
huachaohuang, line 30: The "engula" prefix is not necessary here.

src/kernel/src/grpc/kernel.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::sync::Arc;
+
+use engula_journal::{grpc as grpc_journal, Error as JournalError, Journal};
+use engula_storage::{grpc as grpc_storage, Error as StorageError, Storage};
+use futures::StreamExt;
+
+use super::{client::Client, proto::*};
+use crate::{async_trait, Error, KernelUpdate, Result, ResultStream, Sequence};
+
+const DEFAULT_STREAM: &str = "DEFAULT-STREAM";
```
huachaohuang, line 24: How about just "DEFAULT"? It is cleaner.

src/kernel/src/grpc/kernel.rs
```diff
@@ -0,0 +1,103 @@
...
+        };
+        self.client.apply_update(input).await?;
+        Ok(())
+    }
+
+    /// Returns the current version.
+    async fn current_version(&self) -> Result<Arc<Version>> {
+        let input = CurrentVersionRequest {};
+        let resp = self.client.current_version(input).await?;
+        let version = resp.version.unwrap();
```
huachaohuang, line 88: Can we handle this more nicely? Returns an error if the response is invalid?

src/kernel/src/grpc/kernel.rs
```diff
@@ -0,0 +1,103 @@
...
+        Ok(Arc::new(version))
+    }
+
+    /// Returns a stream of version updates since a given sequence (inclusive).
+    async fn version_updates(&self, sequence: Sequence) -> ResultStream<Arc<VersionUpdate>> {
+        let input = VersionUpdatesRequest { sequence };
+        match self.client.version_updates(input).await {
+            Ok(output) => Box::new(output.map(|result| match result {
+                Ok(resp) => Ok(Arc::new(resp.version_update.unwrap())),
```
huachaohuang, line 97: ditto

src/kernel/src/grpc/kernel.rs
```diff
@@ -0,0 +1,103 @@
...
+                Err(status) => Err(Error::from(status)),
```
huachaohuang, line 98:
```suggestion
                Err(status) => Err(status.into()),
```

src/kernel/src/grpc/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod client;
+mod error;
+mod kernel;
```
huachaohuang, line 17: I think we need to export the `kernel::Kernel` and `server::Server` here. But I think you will use these in another PR. So you can do that later.

src/kernel/src/grpc/proto.rs
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+tonic::include_proto!("engula.kernel.v1");
+
+pub use crate::{Version, VersionUpdate};
```
huachaohuang, line 17: Maybe move this above.

src/kernel/src/grpc/server.rs
```diff
@@ -0,0 +1,78 @@
...
+        request: Request<VersionUpdatesRequest>,
+    ) -> Result<Response<Self::VersionUpdatesStream>, Status> {
+        let input = request.into_inner();
+        let updates_stream = self.kernel.version_updates(input.sequence).await;
+        Ok(Response::new(Box::new(updates_stream.map(
+            |result| match result {
+                Ok(version_update) => Ok(VersionUpdatesResponse {
+                    version_update: Some((*version_update).clone()),
+                }),
+                Err(e) => Err(Status::from(e)),
```
huachaohuang, line 74:
```suggestion
                Err(e) => Err(e.into()),
```

huachaohuang (commented):
Generally looks good to me. I think you will send another PR to add the binary, right?

src/kernel/src/grpc/kernel.rs
PatrickNicholas: How about adding an `Error::Internal` to indicate this type of error?

huachaohuang: Sounds good to me.

PatrickNicholas:
@huachaohuang 
I will send another PR to add the binary and export `Kernel` and `Server`. The above comments have been solved.

tisonkun:
@PatrickNicholas you can edit the PR title to remove the `(#148)` part and write

```
This closes #148.
```

In the PR description so that GitHub [magic](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) takes effect.

src/kernel/src/grpc/kernel.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[derive(Clone)]
+pub struct Kernel {
+    client: Client,
+    journal: grpc_journal::Journal,
+    storage: grpc_storage::Storage,
+}
+
+impl Kernel {
+    #[allow(dead_code)]
```
tisonkun, line 35: Please try to add a test for starting a server and connect to it instead of `#[allow(dead_code)]`.

src/kernel/src/grpc/server.rs
```diff
@@ -0,0 +1,78 @@
...
+
+use super::proto::*;
+use crate::{Kernel, KernelUpdate};
+
+pub struct Server<K: Kernel> {
+    kernel: K,
+}
+
+impl<K: Kernel> Server<K> {
+    #[allow(dead_code)]
```
tisonkun, line 26: ditto

src/kernel/src/grpc/server.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn new(kernel: K) -> Self {
+        Server { kernel }
+    }
+
+    #[allow(dead_code)]
```
tisonkun, line 31: ditto

PatrickNicholas:
> @PatrickNicholas you can edit the PR title to remove the (#148) part and write

@tisonkun Thinks for you suggestions, it works. The `dead_code` have already removed by adding test.

src/kernel/src/grpc/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    async fn test() -> std::result::Result<(), Box<dyn std::error::Error>> {
+        let listener = TcpListener::bind("127.0.0.1:0").await?;
+        let local_addr = listener.local_addr()?;
+        tokio::task::spawn(async move {
+            let j = MemJournal::default();
+            let stream = j.create_stream("DEFAULT").await.unwrap();
+            let s = MemStorage::default();
+            let bucket = s.create_bucket("DEFAULT").await.unwrap();
+            let manifest = mem::Manifest::default();
+            let kernel = mem::Kernel::init(stream, bucket, manifest).await.unwrap();
```
huachaohuang, line 47: The test is very weird here and I think it is because the kernel service and the journal/storage service are not orthogonal. You wrap a memory kernel in a grpc service, but its stream and bucket are not used in this case because the grpc kernel actually connects to a journal/storage service for that. Let me think about it 🤔

huachaohuang: I see, this is actually my fault. I should take a journal/storage directly in other kernel implementations. Then you don't need to manually create a stream/bucket outside. I think we can keep it this way for now and see how to fix it later.

src/kernel/src/grpc/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        let b = kernel.bucket().await?;
+        let mut w = b.new_sequential_writer("object").await?;
+        let buf = vec![0, 1, 2];
+        w.write(&buf).await?;
+        w.flush().await?;
+        w.shutdown().await?;
+        let mut r = b.new_sequential_reader("object").await?;
+        let mut got = Vec::new();
+        r.read_to_end(&mut got).await?;
+        assert_eq!(got, buf);
```
huachaohuang, line 93: I think these tests are unnecessary because they simply test the memory implementation again.

huachaohuang (commented):
Rest LGTM

huachaohuang:
BTW, @tisonkun taught me not to force-push [here](https://github.com/engula/engula/pull/153#issuecomment-984272081). I think he's right and you should try it :)

src/kernel/src/grpc/mod.rs
PatrickNicholas: If testing stream and bucket is unnecessary here, I can change the code to `Kernel::open`.

PatrickNicholas:
> BTW, @tisonkun taught me not to force-push [here](https://github.com/engula/engula/pull/153#issuecomment-984272081). I think he's right and you should try it :)

I see. But I have to force-push again to update the commit message, because I need to change `closes #xxx` to `resolve #xxx`.

PatrickNicholas:
I have not found a way to automatically link an issue and not automatically close the issue when merging that PR. Is any suggestions here?

huachaohuang:
> I have not found a way to automatically link an issue and not automatically close the issue when merging that PR. Is any suggestions here?

What do you mean by "automatically link an issue"? If you mention the issue anywhere, it will show up in the issue.

huachaohuang (approved):
Thanks, LGTM. I think you do a good job here :)

Issue #128, opened 2021-11-22
A file journal implementation


huachaohuang:
We have a `mem` and `grpc` implementation now. But most applications will need persistent journal storage. A sophisticated implementation is not easy, we can start with a naive implementation for now. Check [RocksDB LogWriter](https://github.com/facebook/rocksdb/blob/main/db/log_writer.h) for an example.

ztelur:
Hi, I am interested in this job, plz assign it to me

huachaohuang:
@ztelur Thanks for your interest! Since you are new here, I need to make sure that you have enough context about the job before assigning it to you. So would you like to describe your plan for this job first? Do you need more information?

ztelur:
ok. I am a rookie in storage system. I will make a plan and an approximate timetable

huachaohuang:
@ztelur A simple description about what and how you are going to implement should be good enough. I just need to make sure that we are on the same page.

tisonkun:
@ztelur the abstraction of Journal is defined by @huachaohuang already.

https://github.com/engula/engula/blob/00abae94aa27a520ef1dfd9cdc60b716822f8c13/src/journal/src/journal.rs#L19-L28

...and [a storage implementation based on local files](https://github.com/engula/engula/tree/main/src/storage/src/file) can be referred also.

> I am a rookie in storage system.

I don't think it makes any sense to move this issue forward. If you need more information, spell it out.

DCjanus:
Should we limit size of each event, since this may affect the underlying design.

huachaohuang:
> Before anyone start this task, should we limit size of each event? since this may affect the underlying design and leak from API design.
> 
> For example, for some WAL design, log would not be able to cross fix sized chunk or file, which required most log have a suitable size.

I think the semantic of Journal doesn't specify any size limit. So it is really up to the implementation. But I think a reasonable size range should be a few bytes to tens of megabytes. Of course, someone may choose to implement a Journal that only accepts fixed size events but performs so well that some user just accepts the size constraint.

huachaohuang:
If we assume that there will not be too many streams, then we can map a stream to a sequence of fixed-size files ordered by timestamps. All writes go to the last active file until the file size reaches a specific threshold. Then the last file should be sealed and a new file is created. The idea is very similar to RocksDB WAL.

ygf11:
Hi everyone, I'm also interested in this topic.(@ztelur feel free to take this task).  

[...]

PR #175, opened 2021-12-08
journal: a file journal implementation

ztelur:
Make a draft pr for review
Check #128  for more details.
There is a unit test failure when write and read big event in `mod.big()`. I will fix lately


src/journal/src/file/stream.rs
```diff
@@ -0,0 +1,422 @@
...
+
+        let total_size = read_file.metadata().await?.len();
+
+        let writer = BufWriter::new(write_file);
+        let reader = BufReader::new(read_file);
+
+        let segment = Segment {
+            path: path.clone(),
+            writer: Arc::new(Mutex::new(writer)),
+            reader: Arc::new(Mutex::new(reader)),
```
zojw, line 143: maybe we can only keep the tailer segment's writer  long-lived opening

huachaohuang (commented):
Thanks for your work. I know you have put a lot of effort into this, but I have to say the code quality still needs a lot of improvement. Some advice here:

1. Learn more Rust
2. Improve modularity and capsulation, especially the `Segment` thing
3. Extract the codec thing to one place
4. Avoid unnecessary `str` to `Path` covertions
5. Avoid checking `Option` in different places and in different ways
6. Use better variable/function names

src/journal/src/stream.rs
```diff
@@ -39,6 +39,12 @@ impl From<u64> for Timestamp {
     }
 }
 
+impl fmt::Display for Timestamp {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
```
tisonkun, line 46: I don't think these lines are necessary. We already derive `Debug` for `Timestamp` and you should format it with `{:?}`.

tisonkun: Even if you want `Display` elsewhere, deriving it instead of writing this code is preferred.

src/journal/src/file/error.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn from(e: std::io::Error) -> Self {
+        Error::Unknown(e.to_string())
+    }
+}
+
+impl From<&str> for Error {
+    fn from(s: &str) -> Self {
+        Error::Unknown(s.to_string())
+    }
+}
```
tisonkun, line 29: A conversion from `&str` to `Error` is weird. Why do you need it?

src/journal/src/file/journal.rs
```diff
@@ -0,0 +1,122 @@
...
+        match fs::DirBuilder::new().recursive(true).create(&path).await {
+            io::Result::Ok(_) => {
+                let journal = Journal {
+                    root: path,
+                    streams: Arc::new(Mutex::new(HashMap::new())),
+                };
+                journal.init().await?;
+                Ok(journal)
+            }
+            io::Result::Err(e) => Err(Error::Unknown(e.to_string())),
```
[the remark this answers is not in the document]
ztelur: Done

src/journal/src/file/error.rs
ztelur: it's wrong usage

huachaohuang (commented):
Thanks for your work again. But I'm sorry that I may not consider accepting it. I am not comfortable with the coding style and I would need to comment on half of the changes for review.

ztelur:
OK, looking forward to your code for learn

PR #177, opened 2021-12-09
engula: add cmd to run a kernel server

### LINKED ISSUE #148: A grpc kernel that stores everything in a remote grpc kernel service

PatrickNicholas:
This closes #148.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,7 +136,11 @@ dependencies = [
  "engula-journal",
  "engula-kernel",
  "engula-storage",
+ "futures",
+ "thiserror",
  "tokio",
+ "tokio-stream",
+ "tonic",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,3 +29,7 @@ engula-storage = { path = "src/storage" }
 
 clap = "3.0.0-beta.5"
 tokio = { version = "1", features = ["full"] }
+thiserror = "1.0"
+futures = "0.3"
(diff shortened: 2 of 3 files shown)


PatrickNicholas:
There are alot of error transform codes in this PR. Shall we introduce some libraries (such as anyhow) to solve this problem?

huachaohuang (commented):
Thanks. I can see that there are some dirty things here. So give me some time to refactor and come back to this later.

tisonkun:
In this implementation, we can only launch all units on one node in a server. It may be OK for v0.2 but I think the goal is to have the flexibility to launch arbitrary units.

PatrickNicholas:
> In this implementation, we can only launch all units on one node in a server. It may be OK for v0.2 but I think the goal is to have the flexibility to launch arbitrary units.

I the previous discussion: https://github.com/engula/engula/pull/171#issuecomment-988472342, we decided to serve version 2.0 first. Of course, our goal in the end must be to support launching arbitrary units.

PatrickNicholas:
> So give me some time to refactor and come back to this later

@huachaohuang Is there any updates?

huachaohuang:
> @huachaohuang Is there any updates?

I will come back to this in one or two days. Still working on other stuff.

huachaohuang:
@PatrickNicholas Do you think we should close this in favor of #193 ?

PatrickNicholas:
Now I close this because #193 is merged.

PR #178, opened 2021-12-09
engine: add engine trait and hide kernel from users

huachaohuang:
This PR does two things:

- Add an `Engine` trait to allow different implementations. For example, a remote gRPC engine can be added later.
- Hide `Kernel` from engine users. Each `Engine` implementation can decide which `Kernel` to use internally. When necessary, a dynamic `Engine` implementation can be added, which accepts some configurations to create a demanded engine.

See also some background discussions from [Zulip](https://engula.zulipchat.com/#narrow/stream/306467-general/topic/release.200.2E2/near/264143701).


tisonkun:
Thanks for preparing this PR! I'll review this.

examples/hash_engine.rs
```diff
@@ -12,15 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use engula::{
-    engine::hash::{Engine, Result},
-    kernel::mem::Kernel,
-};
+use engula::engine::hash::{Engine, MemEngine, Result};
```
tisonkun, line 15: I think our ci task doesn't verify examples. You can add a `--all-targets` arg for `cargo test`.

src/engine/hash/src/local/error.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright 2021 The Engula Authors.
```
tisonkun, line 1: Aren't these conversions already covered by:

```rust
    pub fn unknown<E: std::error::Error + Send + 'static>(err: E) -> Self {
        Self::Unknown(Box::new(err))
    }
```

tisonkun (commented):
Generally looks good. Comments inline.

Also, if the engine is the final user-facing interface and it already direct talk to journal(stream) and storage(bucket), letting the kernel hold these two fields only to return them is obsolete.

huachaohuang:
I still feel something doesn't add up, I need more time to think over it.

PatrickNicholas:
> I still feel something doesn't add up, I need more time to think over it.

@huachaohuang  Could you open a discussion to discuss this issue?

huachaohuang:
> > I still feel something doesn't add up, I need more time to think over it.
> 
> @huachaohuang Could you open a discussion to discuss this issue?

Here it is: https://github.com/engula/engula/discussions/182

huachaohuang:
According to the [discussion](https://github.com/engula/engula/discussions/182), I think this is not the direction we are taking, so let's close it.

PR #183, opened 2021-12-10
cargo: following Cargo toml conventions

### LINKED ISSUE #181: Consider following the Cargo.toml conventions
There is a Cargo.toml style guide [here](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/cargo.md). We can consider following it.

ygf11:
Resolves #181

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,31 +1,33 @@
 [package]
 name = "engula"
 version = "0.2.0"
-edition = "2021"
-license = "Apache-2.0"
-homepage = "https://engula.io"
-repository = "https://github.com/engula/engula"
-keywords = ["database", "serverless", "storage-engine"]
 categories = [
   "asynchronous",
   "data-structures",
   "database",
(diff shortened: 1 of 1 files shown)


tisonkun:
@ygf11 I'd suggest you try to reach a consensus on the issue first instead of rushing into a pull request. BTW, is there a check to guard this change? Otherwise incoming patches can break this work easily.

ygf11:
> I'd suggest you try to reach a consensus on the issue first instead of rushing into a pull request. BTW, is there a check to guard this change? Otherwise, incoming patches can break this work easily.

@tisonkun  Sure. Do we discuss here or in the issue?

tisonkun:
Let's go back to the issue.

huachaohuang:
According to https://github.com/engula/engula/issues/181, we should use a tool for this instead of doing it manually. @ygf11 Feel free to send another PR if you figure it out :)

PR #199, opened 2021-12-16
ci: always run cargo install cargo-audit

### LINKED ISSUE #188: Figure out how cargo-cache in GitHub workflow works
@huachaohuang @Xuanwo it seems we just cannot skip the install step even if cargo cache hit.

Cargo cache can already speed up the installation by `cargo-install` cache hit logic, we should always run `cargo install` to avoid false positive.

Remove `if: steps.cache.outputs.cache-hit != 'true'` lines should work. @Xuanwo we may later figure why current settings create false positive.

_Originally posted by @tisonkun in https://github.com/engula/engula/pull/185#issuecomment-992226207_

tisonkun:
`cargo` already handles cache if there is one.

This closes #188.

Verified
* cache miss: https://github.com/tisonkun/engula/runs/4541804122?check_suite_focus=true
* cache hit: https://github.com/tisonkun/engula/runs/4541832408?check_suite_focus=true

### DIFF
--- a/.github/workflows/audit-security.yml
+++ b/.github/workflows/audit-security.yml
@@ -32,7 +32,6 @@ jobs:
   audit-security:
     name: Audit Security
     runs-on: ubuntu-latest
-    if: "!contains(github.event.head_commit.message, 'skip audit')"
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


tisonkun:
cc @mati865 @huachaohuang

huachaohuang (approved):
Thanks

PR #201, opened 2021-12-16
Add some documents and tiny fixes

huachaohuang:
Well, I think this may be the last PR for v0.2.


PatrickNicholas (approved):

tisonkun (approved):

tisonkun:
@huachaohuang I'd like to know how you format `.proto` files so that I can set it locally to avoid further format diff if any.

huachaohuang:
> @huachaohuang I'd like to know how you format `.proto` files so that I can set it locally to avoid further format diff if any.

I am using the [vscode-proto3](https://marketplace.visualstudio.com/items?itemName=zxh404.vscode-proto3) extension with default configurations. I think the format style is identical to clang-format, which is widely used in the C++ world.

PR #204, opened 2021-12-16
cargo: specify package versions and update some packages

huachaohuang:
We need to specify package versions so that users can use a package from crates.io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,9 +40,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.51"
+version = "0.1.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


tisonkun (commented):
May I ask for the rationale for example a link to what we have to do so?

huachaohuang:
The cargo doc [here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) specifies this explicitly. I also tried it and published the 0.2.0-alpha.0 [here](https://crates.io/crates/engula).

huachaohuang:
Note that our code has to be compiled with a nightly toolchain. So if you want to try it, you will need:

```
cargo +nightly run
cargo +nightly install engula --version 0.2.0-alpha.0
```

huachaohuang:
@tisonkun Any update? I need to move forward here.

tisonkun (approved):
Go ahead!

PR #207, opened 2021-12-17
Update README and bump versions

huachaohuang:
The release branch uses the current version number and the main branch should use the next version number.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "engula"
-version = "0.2.0"
+version = "0.3.0"
 edition = "2021"
 license = "Apache-2.0"
 homepage = "https://engula.io"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -22,10 +22,10 @@ path = "bin/engula.rs"
 members = ["src/engine/hash", "src/kernel", "src/journal", "src/storage"]
 
 [dependencies]
-hash-engine = { version = "0.2", path = "src/engine/hash" }
-engula-kernel = { version = "0.2", path = "src/kernel" }
-engula-journal = { version = "0.2", path = "src/journal" }
-engula-storage = { version = "0.2", path = "src/storage" }
+hash-engine = { version = "0.3.0", path = "src/engine/hash" }
```
tisonkun, line 25: I'm curious why `0.2` and `0.3.0` instead of `0.3`?

tisonkun (approved):
LGTM. One question inline.

Cargo.toml
huachaohuang: Oh, good question. Let me see which way is better here.

huachaohuang: I think `x.y.z` is more accurate here, in case we need to release a patch release. I checked some other projects and they all specify `x.y.z`. So I think we can follow them unless we have reasons not to.

tisonkun: Good to know. I should learn from how crates.io works and cooperate on the release manual. It's quite a different model than maven central which I'm familiar with, where a "x.y.z-SNAPSHOT" is preferred here.

How cargo makes an alpha release? What about x.y.z-alpha.0? (Or we can discuss later in the release manual PR)

huachaohuang: To be honest, I don't know all those stuff very much too. I just tried to learn how other projects do that. Maybe we should spend some time reading [the cargo reference](https://doc.rust-lang.org/cargo/reference/publishing.html).

huachaohuang: Oh, cargo actually specifies the rules [here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field).

PR #217, opened 2021-12-21
luna: add table/block builder

huachaohuang:
Use a simplified encoding from LevelDB/RocksDB. For example, things like prefix compression are not implemented.


PatrickNicholas (approved):

tisonkun (approved):
LGTM. We can have a tracking issue for luna engine and ref these PRs to that issue.

PR #223, opened 2021-12-23
bin: support parse localhost

### LINKED ISSUE #222: feature request - Support `localhost` in address
## Behavior

```
$ engula journal run localhost:8888 --mem
error: Invalid value for '<ADDR>': invalid IP address syntax
```

##  Expected

Just works as `127.0.0.1:8888`

##  Version

6c59de2

tisonkun:
This closes #222.

### DIFF
--- a/bin/engula.rs
+++ b/bin/engula.rs
@@ -22,15 +22,17 @@ use engula_kernel::grpc::{FileKernel, MemKernel, Server as KernelServer};
 use engula_storage::{
     file::Storage as FileStorage, grpc::Server as StorageServer, mem::Storage as MemStorage,
 };
+use tokio::net::TcpListener;
+use tokio_stream::wrappers::TcpListenerStream;
 
 pub type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
-macro_rules! run_until_asked_to_quit {
+macro_rules! bootstrap_service {
     ($addr:expr, $server:expr) => {{
-        let cloned_addr = $addr.clone();
+        let listener = TcpListener::bind($addr).await?;
         tonic::transport::Server::builder()
             .add_service($server.into_service())
-            .serve(cloned_addr)
+            .serve_with_incoming(TcpListenerStream::new(listener))
             .await?;
     }};
 }
@@ -52,7 +54,7 @@ enum StorageCommand {
     #[clap(about = "Run a storage server")]
     Run {
         #[clap(about = "Socket address to listen")]
-        addr: SocketAddr,
+        addr: String,
 
         #[clap(subcommand)]
         cmd: RunMode,
@@ -66,11 +68,11 @@ impl StorageCommand {
                 RunMode::File { path } => {
                     let storage = FileStorage::new(&path).await?;
                     let server = StorageServer::new(storage);
-                    run_until_asked_to_quit!(addr, server);
+                    bootstrap_service!(addr, server);
                 }
                 RunMode::Mem => {
                     let server = StorageServer::new(MemStorage::default());
-                    run_until_asked_to_quit!(addr, server);
+                    bootstrap_service!(addr, server);
                 }
             },
         }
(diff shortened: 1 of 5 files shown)


tisonkun:
We may need an e2e test for this kind of changes though.

src/kernel/src/lib.rs
```diff
@@ -49,6 +49,7 @@ pub type ResultStream<T> = Box<dyn futures::Stream<Item = Result<T>> + Send + Un
 pub use self::{
     error::{Error, Result},
     kernel::{Kernel, KernelUpdate},
+    manifest::Manifest,
```
huachaohuang, line 52: Why do we need to export this?

huachaohuang (commented):
Thanks for fixing this issue.

src/kernel/src/lib.rs
tisonkun: It's an intermediate change. Let me revert it.

tisonkun: Updated. I've re-triggered CI due to a weird error message ["compiler panic"](https://github.com/engula/engula/runs/4619961728?check_suite_focus=true).

tisonkun:
@Xuanwo I do think we need to investigate a better workflow now, if you make it clear in https://github.com/datafuselabs/databend/discussions/3415  :P

bin/engula.rs
```diff
@@ -22,15 +22,17 @@ use engula_kernel::grpc::{FileKernel, MemKernel, Server as KernelServer};
...
 
-macro_rules! run_until_asked_to_quit {
+macro_rules! bootstrap_service {
     ($addr:expr, $server:expr) => {{
-        let cloned_addr = $addr.clone();
+        let listener = TcpListener::bind($addr).await?;
         tonic::transport::Server::builder()
             .add_service($server.into_service())
-            .serve(cloned_addr)
+            .serve_with_incoming(TcpListenerStream::new(listener))
```
PatrickNicholas, line 35: What is the difference between `TcpListenerStream` and direct invoke `serve`.

bin/engula.rs
```diff
@@ -127,7 +129,7 @@ enum KernelCommand {
     #[clap(about = "Run a kernel server")]
     Run {
         #[clap(about = "Socket address to listen")]
-        addr: SocketAddr,
+        addr: String,
```
PatrickNicholas, line 132: The journal and storage address is also `SocketAddr`, also need to modify it to `String` to support hostname.

PatrickNicholas (commented):
@tisonkun Hi, thanks for preparing this PR, I did not consider that `SocketAddr` does not support parsing hostname. I still have some comments here.

bin/engula.rs
zojw: it's purpose should be use [TcpListener::bind](https://docs.rs/tokio/1.15.0/tokio/net/struct.TcpListener.html#method.bind) that accept `ToSocketAddrs`, so it can accept both ip or hostname...

> PS: it's little strange feelling to bind a hostname with one of multiple IPs :)

PatrickNicholas: @zojw I got it, thinks.

bin/engula.rs
tisonkun: Yes. I update them at:

* https://github.com/engula/engula/pull/223/files#diff-bb7497ee2ca6fd7abea4b36d3c6a7e38a9fdb2b86717696455597b487baff41eR89
* https://github.com/engula/engula/pull/223/files#diff-bb7497ee2ca6fd7abea4b36d3c6a7e38a9fdb2b86717696455597b487baff41eR57

tisonkun:
> I did not consider that `SocketAddr` does not support parsing hostname.

No, it doesn't. See also the original issue #222.

bin/engula.rs
tisonkun: > it's little strange feelling to bind a hostname with one of multiple IPs

@zojw can you elaborate the sentence? I don't get the point (what is strange).

bin/engula.rs
tisonkun: Oops. I notice that you mean the arguments of this `Command`. It's my mistake. Let me follow it up and try to add a test.

PR #224, opened 2021-12-23
luna: add iterators

huachaohuang:
Add `Iter`, `MergintIter`, and `TwoLevelIter`. We need to add a `IterExt` later. These things work similarly to `AsyncRead` and `AsyncReadExt` in the futures crate.

### DIFF
--- a/src/engine/luna/src/table/iter.rs
+++ b/src/engine/luna/src/table/iter.rs
@@ -1,0 +1,77 @@
+// Copyright 2021 The Engula Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::{
+    ops::DerefMut,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use crate::Result;
+
+pub type BoxIter = Pin<Box<dyn Iter>>;
+
+pub trait Iter {
+    fn poll_seek_to_first(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<()>>;
+
+    fn poll_seek(self: Pin<&mut Self>, cx: &mut Context<'_>, target: &[u8]) -> Poll<Result<()>>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<()>>;
+
+    fn current(&self) -> Option<(&[u8], &[u8])>;
+}
+
+impl<I: ?Sized + Iter + Unpin> Iter for Box<I> {
+    fn poll_seek_to_first(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<()>> {
+        Pin::new(&mut **self).poll_seek_to_first(cx)
+    }
+
+    fn poll_seek(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        target: &[u8],
+    ) -> Poll<Result<()>> {
+        Pin::new(&mut **self).poll_seek(cx, target)
+    }
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<()>> {
+        Pin::new(&mut **self).poll_next(cx)
+    }
+
+    fn current(&self) -> Option<(&[u8], &[u8])> {
+        (**self).current()
+    }
+}
+
+impl<P> Iter for Pin<P>
+where
+    P: DerefMut + Unpin,
+    P::Target: Iter,
+{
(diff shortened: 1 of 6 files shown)


src/engine/luna/src/table/iter.rs
```diff
@@ -0,0 +1,77 @@
...
+    ops::DerefMut,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use crate::Result;
+
+pub type BoxIter = Pin<Box<dyn Iter>>;
+
+pub trait Iter {
```
PatrickNicholas, line 25: Why not use async_trait directly?

huachaohuang: `async_trait` methods will result in a heap allocation per-function-call, which is not acceptable for iterators.

Ref: 

- https://rust-lang.github.io/async-book/07_workarounds/05_async_in_traits.html
- https://docs.rs/async-trait/latest/src/async_trait/expand.rs.html#153

skyzh: But do we really need to wrap a trait for all iterators? If not, we don't need async trait, and we can write them in an async way :)

skyzh: ...in the mean time, we also have https://github.com/tiannian/async-trait-static

PatrickNicholas:
@huachaohuang I have submitted https://github.com/huachaohuang/engula/pull/3 to add `BlockIter`, PTAL.

src/engine/luna/src/table/iter.rs
huachaohuang: > But do we really need to wrap a trait for all iterators?

I think it is necessary to support compositions like `MergintIter` and `TwoLevelIter`.

> If not, we don't need async trait, and we can write them in an async way :)

Do you mean implementing the same set of methods without defining a trait?

> ...in the mean time, we can also use https://github.com/tiannian/async-trait-static

The project seems unmaintained.

huachaohuang:
> @huachaohuang I have submitted [huachaohuang#3](https://github.com/huachaohuang/engula/pull/3) to add `BlockIter`, PTAL.

@PatrickNicholas Thanks, I still need to improve the interfaces a little bit.

src/engine/luna/src/table/iter.rs
skyzh: > Do you mean implementing the same set of methods without defining a trait?

Yes. And we can wrap an enum on all iterators, which is nearly the same as using a trait.

skyzh: ... the only limitation is that there shall be no recursive call in such enum. Otherwise we still need to use boxed future by importing async_recursion crate.

skyzh: Some recursions might not be easy to spot. Anyway, the compiler will complain on such situations.

huachaohuang: What's the benefit of using enum instead of trait?

PatrickNicholas: > What's the benefit of using enum instead of trait?

We would write each iterator in asyncness, and no use async_trait.

huachaohuang: I can't see how that works. Maybe you can submit an alternative implementation of this PR using "the enum way"?

src/engine/luna/src/table/merging_iter.rs
```diff
@@ -0,0 +1,116 @@
...
+                self.buffer.push(iter);
+            }
+        }
+        self.heap = BinaryHeap::from(heap);
+        Poll::Ready(Ok(()))
+    }
+
+    fn poll_seek(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
```
zojw, line 72: after some learning..the semantic for merge seek should be "seek key for all children"~

but maybe here have a problem when "child1(ready), child2(pending), child3(ready)", and `ready!` meet child should returned from this method and re-enter again when child2 ready, but `heap` seems will miss child1 iter in this situation

huachaohuang: Aha, I see, the implementation is problematic.

src/engine/luna/src/table/iter.rs
PatrickNicholas: @huachaohuang I tried to write a snippets: https://gist.github.com/PatrickNicholas/9b050ff86e8649bc66033fd2003e445a

cc @skyzh Did I understand your method correctly?

skyzh: Yes, that's exactly as I thought.

skyzh: ... and you may impl Iter as:

```
impl Iter {
  pub async fn seek(&self, xxx) {
    match self {
      Self::Mem(it) => it.seek(xxx).await,
      Self::Block(it) => it.seek(xxx).await
    }
}
```

So that it's nearly the same as async trait.

huachaohuang: Thanks, looks interesting to me. Although I worry about the performance of matching on every function call. It seems that we don't have many compositions, maybe we can statically implement the following types:

- `BlockIter`
- `TableIter`: index block + data block
- `LevelIter`: a vector of table iterators
- `Iter`: a merging iterator of levels

huachaohuang:
@skyzh @PatrickNicholas Thanks for the discussion. I think we can start implementing [these iterators](https://github.com/engula/engula/pull/224#discussion_r775728933) without the async trait first and see how things go. @PatrickNicholas would you like to refactor https://github.com/huachaohuang/engula/pull/3 to remove the async trait? I think `BlockIter` can be implemented without async. You can send an independent PR to this repo.

Issue #228, opened 2021-12-24
Roadmap 0.3 - Refactor Storage interfaces


huachaohuang:
```rust
pub trait Storage {
    type RandomReader;
    type SequentialWriter;

    async fn create_bucket(&self, bucket_name: &str) -> Result<()>;
    async fn delete_bucket(&self, bucket_name: &str) -> Result<()>;
    async fn delete_object(&self, bucket_name: &str, object_name: &str) -> Result<()>;
    async fn new_random_reader(&self, bucket_name: &str, object_name: &str) -> Result<Self::RandomReader>;
    async fn new_sequential_writer(&self, bucket_name: &str, object_name: &str) -> Result<Self::SequentialWriter>;
}
```

In v0.2, `Kernel` returns a `Bucket` to `Engine`, which is not appropriate because `Bucket` allows `Engine` to delete an object without notifying `Kernel`. So the abstraction needs some refactoring. We can merge `Storage` and `Bucket` and let `Storage` only expose some interfaces with simpler semantics, like `RandomRead`/`SequentialWrite`.

xxchan:
`Bucket` used to provide `SequentialReader`. I'm wondering why do we switch to `RandomReader` here?

huachaohuang:
> `Bucket` used to provide `SequentialReader`. I'm wondering why do we switch to `RandomReader` here?

I think we will provide all `RandomReader`/`RandomWriter`/`SequentialReader`/`SequentialWriter` sooner or later. But currently, we only need `RandomReader` and `SequentialWriter` for Luna engine.

huachaohuang:
I think this job is done in some recent PRs related to `Storage`.

PR #232, opened 2021-12-28
storage: refactor with runtime traits

huachaohuang:
Ref: #228 

This PR is based on #231 . Opt-out things for refactoring.


tisonkun:
Do you drop the `bucket` struct temporarily or permanently? Why?

Issue #229, opened 2021-12-24
Roadmap 0.3 - Refactor Journal interfaces


huachaohuang:
```rust
pub trait Journal {
    type StreamReader;
    type StreamWriter;

    async fn create_stream(&self, name: &str) -> Result<()>;
    async fn delete_stream(&self, name: &str) -> Result<()>;
    async fn new_stream_reader(&self, name: &str) -> Result<StreamReader>;
    async fn new_stream_writer(&self, name: &str) -> Result<StreamWriter>;
}
```

Introduce stream reader/writer to separate concerns. Most use cases will not read and write the same stream at the same time. Separating stream reader/writer has another benefit, allowing different compositions of reader and writer.

tisonkun:
> allowing different compositions of reader and writer

If you use associated type, I'm afraid that each instance of `Journal` has fixed composition of reader and writer. For example:

```rust
impl crate::Journal for MemJournal {
    type StreamReader = MemStreamReader;
    type StreamWriter = MemStreamWriter;
    ...
}
```

w41ter-l:
@huachaohuang The `StreamReader` in Journal interfaces in master branch seems that missing the `watch` method?

w41ter-l:
At the same time, the `StreamWriter` interface cannot express the semantics of a leader based journal(which client is responsible for acting as the leader) and provide a mechanism which detecting leader failures .

huachaohuang:
@w41ter-l We need to further refactor the Journal interfaces. Can you foresee what kind of methods do we need in a leader-based journal implementation?

w41ter-l:
@huachaohuang I thought about it for a while, maybe we don't need to add an interface for `StreamWriter`, but block when calling `new_stream_writer` until it win the campaign. And the leader failure detector is hidden in `new_stream_writer`.

huachaohuang:
I am considering some thing like this:

```rust
enum Event {
    Update(...),
    ChangeRole(Role),
}

enum Role {
    Leader,
    Follower,
}

trait Stream {
    type Watcher: Stream<Item = Event>;
    async fn watch(&self) -> Self::Watcher;
}
```

Multiple instances can watch the same stream at the same time. One of them will receive an event `Event::ChangeRole::Leader` and others will receive `Event::ChangeRole::Follower`. If an instance receives the former, it will act as a leader.

w41ter-l:
[...]

PR #234, opened 2021-12-28
journal: separate stream reader/writer

huachaohuang:
Ref: #229 

This PR is based on #232 . Opt-out things for refactoring.


src/journal/src/stream.rs
```diff
@@ -12,48 +12,33 @@
...
-    }
-
-    pub fn deserialize(bytes: Vec<u8>) -> Result<Self> {
-        let bytes: [u8; 8] = bytes
-            .try_into()
-            .map_err(|v| Error::Unknown(format!("malformed bytes: {:?}", v)))?;
-        Ok(Self(u64::from_be_bytes(bytes)))
-    }
-}
+pub trait Timestamp: Eq + Ord + PartialEq + PartialOrd + Clone + Debug + Send + 'static {}
```
tisonkun, line 20: Why do you reintroduce this trait? IIRC we decide to avoid such abstraction in https://github.com/engula/engula/pull/153.

tisonkun: And you'll sooner or later meet the serde problem again.

PatrickNicholas: +1

huachaohuang: The current `Timestamp` is not friendly for the engine implementers because it doesn't allow them to do anything with the `Timestamp`. I will work out the encoding thing later.

tisonkun: Fine. As these changes almost all for luna engine, maybe an overall review can be applied once the full prototype is ready.

tisonkun:
```rust
/// A composite trait for types that may be used with channels.
#[cfg(feature = "bincode")]
pub trait Data : Send+Sync+Any+Serialize+for<'a>Deserialize<'a>+'static { }
#[cfg(feature = "bincode")]
impl<T: Send+Sync+Any+Serialize+for<'a>Deserialize<'a>+'static> Data for T { }
```

I found a code snippet that can help. @huachaohuang if you somehow stabilize the interface, I'm glad to bring back other implementation and make the binary works again.

huachaohuang: Thanks. Let's stick to the memory implementation for a while. I think the interfaces will not be stable until we get a satisfying luna engine implementation.

Issue #235, opened 2021-12-28
Roadmap 0.3 - Refactor Kernel interfaces


huachaohuang:
The v0.2 kernel only supports a default stream/bucket. We should support multiple streams/buckets in v0.3. We also need to refactor the interfaces corresponding with #228 and #229 .

huachaohuang:
Check #203 for more information.

PR #236, opened 2021-12-28
kernel: refactor and support multiple streams/buckets

huachaohuang:
Ref: #235 

This PR is based on #234 .


src/runtime/src/io/random/random_read.rs
```diff
@@ -0,0 +1,92 @@
...
+            let src = &self[pos..end];
+            let dst = &mut buf[0..src.len()];
+            dst.copy_from_slice(src);
+            src.len()
+        } else {
+            0
+        };
+        Poll::Ready(Ok(len))
+    }
+}
```
xxchan, line 92: I think this file is duplicate with `async_read.rs` (and also for `random_read_ext.rs` & `async_read.ext.rs`, is this a mistake?

src/kernel/src/metadata.proto
```diff
@@ -16,16 +16,18 @@ syntax = "proto3";
 
 package engula.metadata.v1;
 
-message Version {
```
tisonkun, line 19: @huachaohuang do you re-add `Version` later?

With the previous APIs, I can connect an engine with kernel by calling `current_version` and know which objects to subscribe, and create a new stream for use.

After this PR, the Kernel APIs becomes:

```rust
async fn new_stream_reader(&self, stream_name: &str) -> Result<Self::StreamReader>;
async fn new_stream_writer(&self, stream_name: &str) -> Result<Self::StreamWriter>;
async fn new_random_reader(&self, bucket_name: &str, object_name: &str) -> Result<Self::RandomReader>;
async fn new_sequential_writer(&self, bucket_name: &str, object_name: &str) -> Result<Self::SequentialWriter>;
```

... which requires the engine knows the name of object and stream in advance.

tisonkun: Although there is `add_streams` and `add_buckets` field in the `KernelUpdate` message, it seems no consumer for them. An engine can do nothing if it receives these messages since there is no direct API for creating new stream or bucket.

tisonkun: If there is pending tasks, you can create issues for them explicitly.

huachaohuang: You can check the `UpdateWriter` and `UpdateReader`. If you want to create a stream/bucket, you can append a `KernelUpdate` to `UpdateWriter`. If you want to subscribe kernel updates, use `UpdateReader`.
Oh, I think `Kernel` needs to create the streams/buckets when it receives a `KernelUpdate`. Let me do that later.

Issue #226, opened 2021-12-24
Add test for engula binary


tisonkun:
This follow up https://github.com/engula/engula/pull/225 can be avoid if we have a test for engula binary.

However, current CI workflows is a bit repeat itself. If I simply add one new, it requires repeatedly installing Cargo toolchain and build the binary again. Thus, I propose to sort out workflows as @Xuanwo's proposal in https://github.com/datafuselabs/databend/discussions/3415.

I'm not working on this now, so feel free to take it over.

Xuanwo:
I will work for it.

tisonkun:
@Xuanwo after some investigation I notice that you should always setup the environment per job. If so, then it seems introducing `needs` doesn't help a lot.

tisonkun:
After notice that we don't need to introduce a "framework" as databend which seems too complex for now, I'm going to integrate the e2e test by myself.

@Xuanwo you can always prepare a pull request or demo and we can review if what you did in databend valuable for our case here. According to the concern above I don't find a good reason to use, though.

tisonkun:
I'm trying to add an e2e test as in [here](https://github.com/engula/engula/runs/4630797465?check_suite_focus=true). However, it seems "localhost" on GitHub Actions Runner cannot be properly parsed. Also, to wait for servers bootstrap in the background is determinately a bit challenging without liveness detection.

Perhaps we will include servers bootstrap in engula operator or docker compose, and then we can add e2e tests with that infrastructure. FYI @zojw @gaocegege.

PR #238, opened 2021-12-29
ci: build and publish image

tisonkun:

Here is a draft PR for building and publish container image of `engula`. Given that binary in main branch is unstable, I make this draft based on v0.2.0 and hope that we can migrate to the main branch later. This is a preview and I'm going to make use of this image to setup a docker-compose stack for running e2e tests.

Try it out:

```
docker pull ghcr.io/tisonkun/engula:latest
docker run -it ghcr.io/tisonkun/engula:latest # <args>
```

Pending issues:

* [x] `docker/login-action` should login by who has the permission to publish to the repository packages, see also [how rust-lang/docker-rust-nightly did it](https://github.com/rust-lang/docker-rust-nightly/blob/312f155d045659692f36b86170a61a7e544320f0/.github/workflows/build.yml#L45-L55).
    Non issue as I notice that `github.actor` of pushing by merging is the merge actor (i.e., a comitter).
* [x] Reconcile nightly publishes and release publishes.
* Multi-platform support and slim the image are possible, but I tend to leave it as improvements.

Reference:

* #226 
* #170


zojw:
how about helping add `target` folder to `.dockerignore` too~

developer usually `cargo build` in local and do some local test then `docker build` do more test in k8s env.

target folder is big and waste time in `COPY .  /build/` in my local laptop, it's 7.5G and 30s

tisonkun:
@zojw good point. Let me push a follow up for it.

docker-compose.yml
```diff
@@ -18,18 +18,20 @@ services:
...
     command: storage run 0.0.0.0:10002 --mem
   kernel:
     image: ghcr.io/tisonkun/engula:nightly
-    ports:
-      - "10003:10003"
+    network_mode: "service:journal"
     depends_on:
       - journal
       - storage
-    command: kernel run 0.0.0.0:10003 --journal journal:10001 --storage storage:10002 --mem
```
tisonkun: With an offline discussion with @zheaoli I make this network hack. However, a clean solution should be separating runtime network config and client side network config - they have different resolutions.

@zheaoli you can start a discussion on the solution and refer to this commit as an evidence.

Zheaoli: > you can start a discussion on the solution and refer to this commit as an evidence.

Sure I will create a new discussion ASAP tomorrow.

tisonkun: A bit details on this commit:

1. containers in docker-compose services have dedicated namespace ([ref](https://docs.docker.com/compose/networking/))
2. the namespace (alias) is container name
3. Engula's Kernel client asks the server for getting the address of journal & storage
4. If (3) happens on host, the address of journal is `journal:10001`, but in host there is not such an alias, and thus transport error occurs.

tisonkun:
@huachaohuang docker image building and publishing logic is ready. The blocker is that binary is opt-out on main branch. You can ping in this thread when the binary compiled again.

tisonkun:
Closed as rebasing on main branch later.

This should close #268, close #226.

PR #240, opened 2021-12-30
runtime: remove duplicate files

xxchan:
`random_read.rs` and `random_read_ext.rs`(introduced in #236) are duplicate with `async_read.rs` and `async_read_ext.rs` (introduced in #231).


huachaohuang (approved):
Thanks!

PR #246, opened 2022-01-03
rfc: add a tiered storage abstraction

huachaohuang:
I will submit another RFC to describe an implementation of this abstraction later.

### DIFF
--- a/docs/rfcs/00000000-template.md
+++ b/docs/rfcs/00000000-template.md
@@ -1,8 +1,8 @@
-# Title
-
-- Status: '[draft|accepted|released]'
-- Discussion:
-- Pull Request:
+---
+status: '[accepted|released|discarded]'
+discussion: https://github.com/engula/engula/discussions/0
+issue: https://github.com/engula/engula/issues/0
+---
 
 <!--
 We don't have a template of the content yet. Generally, you should think of the motivation, the technical design, the test design, and alternatives of the proposal.
--- a/docs/rfcs/00000000-template.md
+++ b/docs/rfcs/00000000-template.md
@@ -1,8 +1,8 @@
----
-status: '[accepted|released|discarded]'
-discussion: https://github.com/engula/engula/discussions/0
-issue: https://github.com/engula/engula/issues/0
----
+# Title
+
+- Status: '[draft|accepted|released]'
+- Discussion:
+- Pull Request:
 
 <!--
 We don't have a template of the content yet. Generally, you should think of the motivation, the technical design, the test design, and alternatives of the proposal.
--- a/docs/rfcs/2022-01-02-tiered-storage.md
+++ b/docs/rfcs/2022-01-02-tiered-storage.md
@@ -1,0 +1,60 @@
+# Tiered Storage
+
+- Status: draft
+- Pull Request: https://github.com/engula/engula/pull/242
+
+## Abstract
+
+This document proposes a tiered storage abstraction. A tiered storage is an object storage that stores data in multiple storage tiers to provide different cost-performance tradeoffs. It supports flexible options to allow users to decide how to store and access their data.
+
+## Semantic
+
+A tiered storage consists of a hot tier and a cold tier. The hot tier offers faster data access with higher storage costs. The cold tier offers slower data access with lower storage costs.
+
+A tiered storage organizes objects in buckets and provides the following interfaces:
+
+```rust
+trait TieredStorage {
+    /// Creates a bucket.
+    fn create_bucket(&self, bucket);
+    /// Deletes a bucket.
+    fn delete_bucket(&self, bucket);
+    /// Lists buckets in the storage.
+    fn list_buckets(&self);
+    /// Returns a description about the given bucket.
+    fn describe_bucket(&self, bucket);
+    /// Lists objects in the given bucket.
+    fn list_objects(&self, bucket);
+    /// Returns a description about the given object.
+    fn describe_object(&self, bucket, object);
+    /// Returns an object reader for random reads.
+    fn new_random_reader(&self, bucket, object, options: &ReadOptions);
+    /// Returns an object reader for sequential reads.
+    fn new_sequential_reader(&self, bucket, object, options: &ReadOptions);
+    /// Returns an object writer for random writes.
+    fn new_random_writer(&self, bucket, object, options: &WriteOptions);
+    /// Returns an object writer for sequential writes.
+    fn new_sequential_writer(&self, bucket, object, options: &WriterOptions);
+}
+
+enum Tier {
+    Hot,
+    Cold,
+}
+
+struct ReadOptions {
+    /// Whether to store data in the hot tier to speed up future reads.
+    warmup_data: bool,
+    /// The preferred tier to read.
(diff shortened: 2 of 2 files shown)


docs/rfcs/00000000-template.md
```diff
@@ -1,8 +1,8 @@
----
-status: '[accepted|released|discarded]'
-discussion: https://github.com/engula/engula/discussions/0
-issue: https://github.com/engula/engula/issues/0
----
+# Title
+
+- Status: '[draft|accepted|released]'
+- Discussion:
```
tisonkun, line 4: A demo link should help the write to understand what is expected to be placed here. BTW, what's the difference between a yaml header and such item list?

docs/rfcs/00000000-template.md
```diff
@@ -1,8 +1,8 @@
...
+- Pull Request:
```
tisonkun, line 5: Ditto

docs/rfcs/00000000-template.md
tisonkun: If there is not an obvious reason, I suggest you create a dedicated PR to update the template so that we can focus on the rfc in this PR.

docs/rfcs/00000000-template.md
```diff
@@ -1,8 +1,8 @@
----
-status: '[accepted|released|discarded]'
-discussion: https://github.com/engula/engula/discussions/0
-issue: https://github.com/engula/engula/issues/0
----
+# Title
+
+- Status: '[draft|accepted|released]'
```
tisonkun, line 3: The reason '[accepted|released|discarded]' into use is that we think of when the proposal is merged into the codebase. A `draft` status should never occur when a pull request is merged. Also, it's possible that we later find the RFC fails to released/implemented and mark it as discarded.

docs/rfcs/2022-01-02-tiered-storage.md
```diff
@@ -0,0 +1,60 @@
+# Tiered Storage
+
+- Status: draft
+- Pull Request: https://github.com/engula/engula/pull/242
```
tisonkun, line 4:
```suggestion
- Pull Request: https://github.com/engula/engula/pull/246
```

typo?

docs/rfcs/2022-01-02-tiered-storage.md
```diff
@@ -0,0 +1,60 @@
+# Tiered Storage
```
tisonkun, line 1: What's the difference between a tiered storage and a r/w cache? Could you draw a graph on the read/write process? The distinguished options seem like a decision when to build cache.

tisonkun: I found https://github.com/engula/engula/discussions/245 and think this is a synonym of a cache. If so, the general design looks good. One comment is that whether we'd like to provide multi-tiers(>2) cache?

docs/rfcs/00000000-template.md
huachaohuang: The YAML header doesn't get rendered in my editor and I don't like the table format it is rendered [here](https://github.com/engula/engula/blob/main/docs/rfcs/00000000-template.md). I think we can do the job with the standard Markdown syntax.

docs/rfcs/2022-01-02-tiered-storage.md
huachaohuang: This RFC simply proposes an abstraction and its semantic. How it reads/writes or whether there is a cache inside is an implementation detail.

huachaohuang: I also considered supporting multiple tiers like hot, warm, and cold. But the semantic will be very vague and it's hard for users to reason about what should be warm data and what the implementation will do with warm data.

tisonkun: What's the output of this RFC then? Generally, an RFC is valuable with an (reference) implementation.

docs/rfcs/00000000-template.md
huachaohuang: I will update the template in another PR.

docs/rfcs/2022-01-02-tiered-storage.md
tisonkun: > I also considered supporting multiple tiers like hot, warm, and cold. But the semantic will be very vague and it's hard for users to reason about what should be warm data and what the implementation will do with warm data.

The point here is that if the design has exact two options which are under the cache semantic, it can be a simple boolean option instead of enum.

huachaohuang: > What's the output of this RFC then? Generally, an RFC is valuable with an (reference) implementation.

A trait will be introduced to reflect this abstraction. I will also send another RFC that describes an implementation of this abstraction.

docs/rfcs/00000000-template.md
tisonkun: Please mark the comments related as resolved to toggle them :) 

It's a bit weird that as the comment author I don't have permissions to resolve the comment I made :(

docs/rfcs/2022-01-02-tiered-storage.md
huachaohuang: > The point here is that if the design has exact two options which are under the cache semantic, it can be a simple boolean option instead of enum.

OK, then I think we can name it a *cached storage*.

huachaohuang: Hmm, I think cache is actually a very common option, so maybe we can simply merge this semantic with the existing storage abstraction.

tisonkun: > Hmm, I think cache is actually a very common option, so maybe we can simply merge this semantic with the existing storage abstraction.

Yes. This looks good. @zhongligao & @PsiACE have talked about `cache` https://github.com/engula/engula/discussions/76 and I think it can be integrated into storage abstraction as an internal layer/concept.

docs/rfcs/2022-01-02-tiered-storage.md
```diff
@@ -0,0 +1,62 @@
...
+    /// Lists objects in the given bucket.
+    fn list_objects(&self, bucket);
+    /// Returns a description about the given object.
+    fn describe_object(&self, bucket, object);
+    /// Returns an object reader for random reads.
+    fn new_random_reader(&self, bucket, object, options: &ReadOptions);
+    /// Returns an object reader for sequential reads.
+    fn new_sequential_reader(&self, bucket, object, options: &ReadOptions);
+    /// Returns an object writer for random writes.
+    fn new_random_writer(&self, bucket, object, options: &WriteOptions);
```
w41ter-l, line 37: Shall we support random writer? That means we wouldn't support a lot of distributed file system which only support appending semantics.

huachaohuang: We can support random read and sequential write first. An implementation doesn't have to support all of these, as long as its target users don't require it to.

huachaohuang:
I will refactor the current storage to provide the similar interfaces described in this document. I think an RFC for this is overkill, we can add some comments in the code instead. So let's close this PR.

PR #247, opened 2022-01-04
rfc: add a shared storage

huachaohuang:
Rendered version: https://github.com/huachaohuang/engula/blob/cached-storage-rfc/docs/rfcs/2022-01-04-cached-storage.md

Ref: #245  #246


tisonkun:
Shall this PR supersedes #246 and thus we can close #246 preferring this one?

huachaohuang:
> Shall this PR supersedes #246 and thus we can close #246 preferring this one?

#246 describes an abstraction and this PR describes a more concrete design. But I have closed #246 anyway.

docs/rfcs/2022-01-04-cached-storage.md
```diff
@@ -0,0 +1,39 @@
...
+
+This RFC proposes a cached storage that caches objects from another object storage to improve read performance. The cached storage manages a set of cache storages with a storage orchestrator.
+
+## Architecture
+
+![Architecture](images/cached-storage-architecture.drawio.svg)
+
+A cached storage consists of a master, a base storage, a set of cache storages, and a storage orchestrator. The base storage is the single point of truth and should offer reliable object storage. The cache storages cache objects from the base storage to improve read performance.
+
+The master serves as a router and a placement driver. It pulls information from all cache storages to form a global view of all objects. Each cache storage collects statistics of objects for the master. The master is then responsible for balancing load among cache storages.
```
zojw, line 16: does master need persistent meta data for itself ? or just all recovery from base storage(s3) metadata after restart node?

zojw (commented):
try to taking into account how to implement and have some detail question :)

docs/rfcs/2022-01-04-cached-storage.md
```diff
@@ -0,0 +1,39 @@
...
+
+To scale the set of cache storages on-demand, the master provisions or de-provisions cache storages from the orchestrator. The orchestrator maintains the list of cache storages for the master. The orchestrator is also responsible for detecting and recovering faulted cache storages.
+
+### Read path
+
+To read an object, a client contacts the master to get a list of locations that serve the object. If the object is not cached, the client reads from the base storage and then notifies the master to cache the object for future reads.
```
w41ter-l, line 22: > a client contacts the master

Will a location cache be added here? Otherwise, the master will become a bottleneck.

> to get a list of locations that serve the object.

It means a object would exists in many cache storage?

docs/rfcs/2022-01-04-cached-storage.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Write path
+
+The write an object, a client contacts the master to get a list of locations to store the object. The client must ensure that the object has been persisted in the base storage before claiming a successful write. The client can further ensure that the object has been cached to avoid reading from the base storage later.
```
w41ter-l, line 26: > The write an object, a client contacts the master to get a list of locations to store the object.

How about reusing the existing mechanism in reading path , instead of a writer contacts to master to get locations to store the object?

> notifies the master to cache the object for future reads.

docs/rfcs/2022-01-04-cached-storage.md
huachaohuang: > Will a location cache be added here? Otherwise, the master will become a bottleneck.

The upper-level modules (`Kernel` and `Engine`) should have their own cache to improve performance.

> It means a object would exists in many cache storage?

Yes, for availability and load balance.

docs/rfcs/2022-01-04-cached-storage.md
huachaohuang: What do you mean "existing mechanism in reading path"?

docs/rfcs/2022-01-04-cached-storage.md
huachaohuang: The master mainly maintains routing information for cache storages, which doesn't need to be persistent. However, we still need to make the master highly available. If we integrate the master into `Kernel`, then we can let the leader kernel instance serve as the storage master too.

docs/rfcs/2022-01-04-cached-storage.md
```diff
@@ -0,0 +1,39 @@
...
+
```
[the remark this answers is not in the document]
huachaohuang: I think option 1 is simpler. Both delete and refill commands go through the master, so it is not hard to do concurrency control.

docs/rfcs/2022-01-04-cached-storage.md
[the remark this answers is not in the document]
huachaohuang: We can have some global options like `min_cache_replicas` and `max_cache_replicas`, and allow the storage to decide the appropriate number of replicas for individual objects for load balance.

w41ter-l: @huachaohuang 
> If the object is not cached, the client reads from the base storage and then notifies the master to cache the object for future reads.

huachaohuang:
I will start prototyping to get more information about the design to further improve the document.

docs/rfcs/2022-01-04-cached-storage.md
```diff
@@ -0,0 +1,39 @@
...
+## Implementation
+
```
ZhongliGao, line 29: If the access data is concentrated on a part of the server, it is easy to cause the underutilization of cloud resources. How should we solve this situation?

docs/rfcs/2022-01-04-cached-storage.md
```diff
@@ -0,0 +1,39 @@
...
+A base storage can be built on a cheap and highly reliable cloud object storage(e.g., AWS S3). A cache storage can be a custom-built storage service that stores data on local SSD or cloud block storage (e.g., AWS EBS). An orchestrator can be built on Kubernetes, which acts as an operator. Kubernetes provides most of the features we need from the orchestrator.
+
```
ZhongliGao, line 31: If the access data is concentrated on a part of the server, it is easy to cause the underutilization of cloud resources. How should we solve this situation?

huachaohuang: Yes, that's why we need the master to do load balancing.

ZhongliGao: > Yes, that's why we need the master to do load balancing.

Is there a specific solution for this load balancing? I'm more interested in this and maybe can do some work on it. And in the paper you referenced, I found that anna19 then optimizes the cloud resource rate for anna18.

huachaohuang: I only get some vague ideas some far. For example, balance load according to access statistics of individual objects. The load balancing strategy doesn't affect the overall architecture, so we may leave that to future works.

docs/rfcs/2022-01-04-cached-storage.md
```diff
@@ -0,0 +1,39 @@
+# Cached Storage
+
+- Status: draft
+- Pull Request:
```
tisonkun, line 4:
```suggestion
- Status: accepted
- Pull Request: https://github.com/engula/engula/pull/247
- Tracking Issue: https://github.com/engula/engula/issues/263
```

tisonkun: co https://engula.zulipchat.com/#narrow/stream/306467-general/topic/random/near/267827815

cc @huachaohuang @zojw

huachaohuang:
This RFC is ready. Although some implementation details are missing, it should be good enough in this early stage to get the work in #263 started.

docs/rfcs/2022-01-04-shared-storage.md
```diff
@@ -0,0 +1,39 @@
...
+
+To read an object, a client contacts the master to get a list of locations that serve the object. If the object is not cached, the client reads from the base storage and then notifies the master to cache the object for future reads.
+
+### Write path
+
+To write an object, a client contacts the master to get a list of locations to store the object. The client must ensure that the object persisted in the base storage before claiming a successful write. The client can further ensure that the object has been cached to avoid reading from the base storage later.
+
+### Implementation
+
+A base storage can be built on a cheap and highly reliable cloud object storage(e.g., AWS S3). A cache storage can be a custom-built storage service that stores data on local SSD or cloud block storage (e.g., AWS EBS). An orchestrator can be built on Kubernetes, which acts as an operator. Kubernetes provides most of the features we need from the orchestrator.
```
skyzh, line 35: If the base storage is S3, at what level do you want to cache? (e.g., SST level, multi-part upload level, block level)

From public benchmarks like https://github.com/dvassallo/s3-benchmark, S3 services' latency is relatively high, and will be affected by multiple factors, like how the file is uploaded to s3, the range of get, whether EC2 accesses S3 using VPC, and the configuration of EC2 itself, etc. From my perspective, apart from cache service design, the cache algorithm and cache content is also an interesting part to investigate.

huachaohuang: Thanks for your information. I'm only considering TP scenarios for now. In this case, I think the basic strategy is to cache all data since we don't want to see even a single read from S3. In the future, we may leave some cool SST in S3 according to the access statistics of individual SST to further save costs. But in the current stage, a simple full cache is good enough for the luna engine.

docs/rfcs/2022-01-04-shared-storage.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Motivation
+
+Most object storage services provide high durability and high throughput. But the latency is about tens or hundreds of milliseconds. It is not a problem for background writes, but it is unacceptable for foreground reads of a real-time database. To reduce the read latency, we can add a cache tier on top of an object storage.
+
+## Detailed design
+
+![Architecture](images/shared-storage-architecture.drawio.svg)
+
+`SharedStorage` consists of a master, a base storage, a set of cache storages, and a storage orchestrator. The base storage is the single point of truth and should offer reliable object storage. The cache storages cache objects from the base storage to improve read performance.
```
tisonkun, line 19: Please define "master", "base storage", "orchestrator" before you talk about them. They're all new concepts without definition.

For example, a master is a coordinator of all `cache storage` instances that lives along with the `Kernel`. A base storage is...what? A storage server or S3 cluster?

You may try to connect this architecture with the [overall design](https://github.com/engula/engula/blob/main/docs/design.md) so that we know what part it is of Engula, instead of an isolated design.

docs/rfcs/2022-01-04-shared-storage.md
```diff
@@ -0,0 +1,39 @@
...
+
+The master serves as a router and a placement driver. It pulls information from all cache storages to form a global view of all objects. Each cache storage collects statistics of objects for the master. The master is then responsible for balancing load among cache storages.
+
+To scale the set of cache storages on-demand, the master provisions or de-provisions cache storages from the orchestrator. The orchestrator maintains the list of cache storages for the master. The orchestrator is also responsible for detecting and recovering faulted cache storages.
+
+### Read path
+
+To read an object, a client contacts the master to get a list of locations that serve the object. If the object is not cached, the client reads from the base storage and then notifies the master to cache the object for future reads.
```
tisonkun, line 27: Pseudo code or ordered list could be better.

docs/rfcs/2022-01-04-shared-storage.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Write path
+
+To write an object, a client contacts the master to get a list of locations to store the object. The client must ensure that the object persisted in the base storage before claiming a successful write. The client can further ensure that the object has been cached to avoid reading from the base storage later.
```
tisonkun, line 31: Pseudo code or ordered list could be better.

tisonkun (commented):
Comments inline for understanding.

tisonkun:
@huachaohuang according to current state, I tend to regard this RFC and #280 as an evolving design & implementation documents instead of a proposal required to be accepted before we start implementing.

In this way, we can continuously prototyping while keep polishing such document to stabilize our ideas, instead of forcing to merge something indeterminate or unclear.

huachaohuang:
> @huachaohuang according to current state, I tend to regard this RFC and #280 as an evolving design & implementation documents instead of a proposal required to be accepted before we start implementing.
> 
> In this way, we can continuously prototyping while keep polishing such document to stabilize our ideas, instead of forcing to merge something indeterminate or unclear.

Sounds good to me. My original intention about this document is just to align the design and guide the development, instead of figuring out all the details first. I think we can leave these PRs as drafts and evolve them with the implementation until they are stable enough.

huachaohuang:
Closed in favor of #361.

PR #249, opened 2022-01-04
futures: add io and stream traits

### LINKED ISSUE #248: Add a crate for common async traits and utilities
We will need some common async traits and utilities as the project develops. We can add a crate like `engula-futures` to host the runtime independent things instead of mixing them in `engula-runtime`.

huachaohuang:
Closes #248 

The runtime crate is removed in favor of this runtime independent crate. We may still bring `engula-runtime` back when we really implement our own runtime.


tisonkun:
Thanks for preparing this PR! It seems more than a simple renaming work. I'm reviewing now.

src/futures/Cargo.toml
```diff
@@ -1,14 +1,14 @@
...
 repository = "https://github.com/engula/engula"
-description = "The Engula asynchronous runtime."
+description = "Asynchronous utilities for Engula."
 
 [dependencies]
 futures = "0.3.19"
 
 [dev-dependencies]
-tokio = { version = "1.15.0", features = ["full"] }
+tokio = "1.15.0"
```
tisonkun, line 14:
```suggestion
tokio = { version = "1.15.0", features = ["full"] }
```

We need `"macros", "rt"` features for running tests individually. Since it's a dev dependency, we can simply set features to `"full"`.

src/futures/src/io/random/read.rs
```diff
@@ -19,7 +19,7 @@ use std::{
     task::{Context, Poll},
 };
 
-pub trait AsyncRead {
+pub trait Read {
     fn poll_read(
```
tisonkun, line 23: You may think of rename this method as `poll_read_at` and emphasize we read from a position as `poll_read_batch` emphasizes we read a batch.

A real-world concern is that `poll_read` name overrides confuses language server (see below). But it's not a blocker anyway.

`poll_read`:

image

`poll_read_at`:

image

src/futures/src/io/random/read_ext.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+        while !this.buf.is_empty() {
+            let n = ready!(Pin::new(&mut this.inner).poll_read(cx, this.buf, this.pos))?;
+            if n == 0 {
+                return Poll::Ready(Err(io::ErrorKind::UnexpectedEof.into()));
+            }
+            let (_, rest) = std::mem::take(&mut this.buf).split_at_mut(n);
+            this.buf = rest;
+            this.pos += n;
```
tisonkun, line 103: @zojw I think https://github.com/engula/engula/pull/231#discussion_r776357618 is resolved by this line :)

src/futures/src/stream/batch_stream.rs
```diff
@@ -0,0 +1,63 @@
...
+use std::{
+    ops::{Deref, DerefMut},
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use futures::Stream;
+
+/// An extended stream that can return a batch of items at once.
+pub trait BatchStream: Stream {
```
tisonkun, line 24: Why `BatchStream` has a bound `futures::Stream`? Shall we implement `futures::Stream.poll_next` from `BatchStream.poll_next_batch`? What's the relationship between `type Item` and `type Batch`?

Maybe you can provide or describe a reference implementation. Otherwise, they're different focus and the user can implement separately.

tisonkun (commented):
Comments inline.

src/futures/src/io/random/read.rs
huachaohuang: I don't have a strong opinion about this. @zojw @w41ter-l What's your opinion?

huachaohuang:
You can check an example usage of some traits defined here at #250 .

src/futures/src/io/random/read.rs
w41ter-l: I agree with `poll_read_at`.

huachaohuang: Hmm, there is another problem, if we use `read_at`, then should we use `read_exact` or `read_at_exact`? In this case, I prefer to simply use `read` and `read_exact`. The trait is specific to random access, so I think there is no confusion here.

tisonkun: @huachaohuang OK. I'll create an issue at intellij-rust repo for the IDE issue dedicated.

huachaohuang: BTW, vscode doesn't have the problem :)

src/futures/src/stream/batch_stream.rs
huachaohuang: I have removed the `futures::Stream`. You can check an example usage at #250 .

w41ter-l (approved):

PR #253, opened 2022-01-05
futures: replace BatchStream with BatchResultStream

huachaohuang:
Check a use case of this stream in #252 .


zojw (dismissed):

src/storage/src/local/mem/storage.rs
```diff
@@ -33,15 +33,14 @@ use crate::{async_trait, Error, Result};
...
 }
 
 impl Default for Storage {
     fn default() -> Self {
         Self {
-            buckets: Arc::new(Mutex::new(HashMap::new())),
+            buckets: Mutex::new(HashMap::new()),
         }
     }
 }
```
tisonkun, line 46: unrelated changes?

huachaohuang: Yeah, I can open another PR for this if necessary.

tisonkun: @huachaohuang It's OK or you can separate in a new commit and use "rebase and commit". I don't insist a new PR but distinguish in commits.

w41ter-l (approved):

Issue #221, opened 2021-12-22
Make a workable Luna engine


huachaohuang:
## Database

- [x] #294
- [x] #297 
- [x] #299 
- [x] #300
- [x] #302
- [x] #303
- [x] #318
- [x] #319
- [x] #321
- [x] #320
- [x] #308
- [x] #310
- [x] #337
- [x] #338
- [ ] Implement compactions
- [x] #343
- [ ] #346

## Table

- [x] #217
- [x] #255 
- [x] #217
- [x] #255 
- [x] #305
- [x] #313

PatrickNicholas:
@huachaohuang I can try to complete the BlockIter after the iterator is defined. Do you have any plans?

huachaohuang:
@PatrickNicholas I haven't got time to make a more clear plan yet. I implemented a part of iterators in https://github.com/engula/engula/pull/224. If you are interested, you can help to review and test. Then you can try to implement things like BlockIter, TableReader, Filter, compressions, etc.

huachaohuang:
BTW, the demo 1 implementation can be used as a reference.

huachaohuang:
`TableReader` requires an asynchronous random access reader. We need to abstract a set of IO traits instead of those from Tokio. For example, `SequentialRead`, `SequentialWrite`, `RandomRead`, `RandomeRead`.

tisonkun:
@huachaohuang I notice that current luna engine doesn't talk to kernel. What's the role the kernel plays for supporting the luna engine?

huachaohuang:
> @huachaohuang I notice that current luna engine doesn't talk to kernel. What's the role the kernel plays for supporting the luna engine?

It will be similar to the hash engine.

tisonkun:
@huachaohuang could you create issues for all subtasks? As in https://github.com/pingcap/tidb/issues/28170, so that I can try to find contributors for taking every subtask and it's more easy to track the process (for now, BlockReader & TableReader are done but not updated) and to understand what tasks are concretely focused.

huachaohuang:
Closed in favor of #359.

PR #255, opened 2022-01-06
luna: add table reader

w41ter-l:
This PR ref #221


huachaohuang:
Thanks, I will take a look later today.

zojw (dismissed):

src/engine/luna/src/table/table_iter.rs
```diff
@@ -0,0 +1,113 @@
...
+            if self.index_iter.valid() {
+                let mut iter = read_block_iter(&mut self.reader, self.index_iter.value()).await?;
+                iter.seek_to_first();
+                self.block_iter = Some(iter);
+            } else {
+                self.block_iter = None;
+            }
+        }
+
+        Ok(())
```
zojw, line 96: BTW: when do we need to return `Err(..)`~? it seems current we need `.valid` after `next` :)

w41ter-l: `next` returns `Err` if it meets IO error. In this function, `read_block_iter` might returns `Err`.

zojw: thanks got it... I miss `?`  orz

src/engine/luna/src/table/table_reader.rs
```diff
@@ -0,0 +1,258 @@
...
+        let index_block_handler = BlockHandle::decode_from(&mut footer.as_slice());
+        let block_content = read_block(&mut reader, &index_block_handler).await?;
+
+        Ok(TableReader {
+            reader,
+            index_block: BlockReader::new(block_content),
+        })
+    }
+
+    pub fn iter(&mut self) -> TableIter<R> {
```
huachaohuang, line 46: I think `TableReader` should be thread-safe and doesn't require a mutable reference to read. I created a related issue [here](https://github.com/engula/engula/issues/261). Maybe you can help to resolve it?

w41ter-l: I'll fix that problem first and then update this PR.

huachaohuang: I think this is not solved?

w41ter-l: Oh, I missed this part.

w41ter-l: @huachaohuang already updated.

huachaohuang (approved):
LGTM. Note that I am not concerned about the code quality in the engine part for now.

PR #257, opened 2022-01-06
journal: refactor traits and remove the timestamp

huachaohuang:
Ref: #229


w41ter-l (approved):

tisonkun (commented):
Do you decide to postpone the topic "user define timestamp"? What's the difference between `Timestamp` and `Sequence`?

tisonkun (approved):

src/journal/src/journal.rs
```diff
@@ -12,13 +12,19 @@
...
-pub trait Journal<T>: Clone + Send + Sync + 'static {
-    type StreamReader: StreamReader<T>;
-    type StreamWriter: StreamWriter<T>;
+pub trait Journal: Send + Sync + 'static {
+    type StreamLister: BatchResultStream<Elem = String, Error = Error>;
+    type StreamReader: StreamReader;
+    type StreamWriter: StreamWriter;
+
+    /// Lists streams.
+    async fn list_streams(&self) -> Result<Self::StreamLister>;
```
tisonkun, line 27: If `list_streams` returns a list of all streams' names, why not just return `Result<Vec<String>>`? It seems unnecessary complexity here.

tisonkun: Since it's a batch poll interface, I can work with this interface now. But still be curious about the quantity of streams of one journal in your mind @huachaohuang.

PR #262, opened 2022-01-06
ci: turn on some taplo rules to format Cargo.toml

xxchan:
As discussed in https://github.com/engula/engula/issues/181, Cargo.toml style guide is not respected. `rustfmt` will support it in the future. 

As a temporary solution, we turn on some `taplo` options, including: 
- sort keys within each section (I think this is the most likely scenario)
- array auto collapse & expand.

### DIFF
--- a/taplo.toml
+++ b/taplo.toml
@@ -1,0 +1,13 @@
+include = ["**/Cargo.toml"]
+
+[formatting]
+array_auto_collapse = true
+array_auto_expand = true
+reorder_keys = true
+
+# Do not sort key names within [package] section.
+[[rule]]
+keys = ["package"]
+
+[rule.formatting]
+reorder_keys = false
--- a/taplo.toml
+++ b/taplo.toml
@@ -1,8 +1,6 @@
 include = ["**/Cargo.toml"]
 
 [formatting]
-array_auto_collapse = true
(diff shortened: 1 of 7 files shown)


tisonkun (approved):
LGTM. Thanks!

cc @huachaohuang

tisonkun:
@xxchan one comment is that whether it's OK to close #181?

taplo.toml
```diff
@@ -0,0 +1,13 @@
+include = ["**/Cargo.toml"]
+
+[formatting]
+array_auto_collapse = true
+array_auto_expand = true
```
huachaohuang, line 5:
```suggestion
```

These two seem to be default configurations, should we remove them?

huachaohuang (commented):
Rest LGTM, thanks!

huachaohuang (approved):

xxchan:
> @xxchan one comment is that whether it's OK to close #181?

Strictly speaking, it's not completely resolved. Personally speaking, current status is good enough and we can just close it to reduce distraction.

huachaohuang:
> Strictly speaking, it's not completely resolved. Personally speaking, current status is good enough and we can just close it to reduce distraction.

Agree, it's good enough, let's close it.

PR #266, opened 2022-01-07
Remove some trait bounds that are not required

huachaohuang:


tisonkun (commented):
Why do we original bring these bounds in fa69337e2179cbc15efd2dbf8a4b0a72cb5952cb?

Is it possible we share references of a journal or storage in a gRPC implementation? For example, hold a reference to the struct in different thread by different method call.

Anyway, we can add the bounds when needed.

zojw (approved):

huachaohuang:
> Why do we original bring these bounds in [fa69337](https://github.com/engula/engula/commit/fa69337e2179cbc15efd2dbf8a4b0a72cb5952cb)?
> 
> Is it possible we share references of a journal or storage in a gRPC implementation? For example, hold a reference to the struct in different thread by different method call.
> 
> Anyway, we can add the bounds when needed.

Maybe some implementations need the trait bound, but not all of them. So I think we should let the specific implementation specify the trait bounds when necessary.

tisonkun (approved):
Thanks for your explanation @huachaohuang. LGTM.

It could reduce one round Q&A if more description provided :)

w41ter-l (approved):

huachaohuang:
> FYI I'm trying to bring back gRPC journal to see if these trait bounds help.

I suggest that we leave the gRPC journal aside for now because the journal API is still unstable as described in #229 and #260.

tisonkun:
@huachaohuang OK. I notice that I can test it on v0.2.0 and confirm that these trait bounds take effect for a gRPC implementation at that version. Possibly we will bring them back later.

And yes, "let the specific implementation specify the trait bounds when necessary", while if the Server takes a type parameter `T: Journal`, I'm afraid we will bring them back.

PR #267, opened 2022-01-07
futures: allow users to bind a batch size to ResultStream

### LINKED ISSUE #91: Find a practice to return asynchronous paged results
We have some APIs that need to return a list of values. For example, get a list of object names from a bucket. The simplest way is to return a `Vec<T>`. But if the list goes too long, it is not a good idea to return it all at once. A common solution for this kind of problem is paging. Rust provides a concept called `Stream`, which is like an asynchronous iterator that allows the caller to get one value at a time.

I think the most mature way to do that is to use [Stream](https://docs.rs/futures/0.3.17/futures/stream/trait.Stream.html) from the `futures` crate. But it is not a good idea for our public APIs to depend on a third-party trait, which is not stable either. Fortunately, Rust is working on a built-in [`Stream`](https://doc.rust-lang.org/std/stream/trait.Stream.html) trait (nightly), so we may consider using that in our public APIs when it is stablized.

huachaohuang:
Closes #91.

Simply speaking, this PR allows users to do

```rust
let stream = stream.batch_size(10);
stream.next();
stream.collect();
```

instead of

```rust
stream.next(10);
stream.collect(10);
```

I think the first usage is handier because most use cases only need to specify the batch size once.

Note that I also update the rust toolchain because the previous one panics on compiling this PR 🤣


src/futures/src/stream/batch/result_stream_ext.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use super::{batch_size::BatchSize, ResultStream};
+
+pub trait ResultStreamExt: ResultStream {
+    fn next(&mut self, batch_size: usize) -> NextFuture<'_, Self>
```
tisonkun, line 24:
```suggestion
    /// Consumes and returns the next `batch_size` values in the stream.
    fn next(&mut self, batch_size: usize) -> NextFuture<'_, Self>
```

src/futures/src/stream/batch/batch_size.rs
```diff
@@ -0,0 +1,53 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::{ResultStream, ResultStreamExt};
+
+pub struct BatchSize<T> {
```
tisonkun, line 17: `BatchSize` doesn't look a good name as it's actually a stream, not a plain data structure.

src/futures/src/stream/batch/batch_size.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    pub fn batch_size(self, batch_size: usize) -> Self {
+        Self {
+            inner: self.inner,
+            batch_size,
+        }
+    }
+
+    pub async fn next(&mut self) -> Result<Vec<T::Elem>, T::Error> {
```
tisonkun, line 38:
```suggestion
    /// Consumes and returns the next `self.batch_size` values in the stream.
    pub async fn next(&mut self) -> Result<Vec<T::Elem>, T::Error> {
```

src/futures/src/stream/batch/batch_size.rs
```diff
@@ -0,0 +1,53 @@
...
+        self.inner.next(self.batch_size).await
+    }
+
+    pub async fn collect(mut self) -> Result<Vec<T::Elem>, T::Error> {
```
tisonkun, line 42:
```suggestion
    /// Drain stream pushing all emitted values into a collection by repeatedly
    /// poll the next `self.batch_size` values.
    pub async fn collect(mut self) -> Result<Vec<T::Elem>, T::Error> {
```

src/futures/src/stream/batch/result_stream.rs
```diff
@@ -18,8 +18,8 @@ use std::{
     task::{Context, Poll},
 };
 
-/// A stream that returns a batch of elements at a time.
-pub trait BatchResultStream {
+/// A stream that yields a batch of elements at a time.
+pub trait ResultStream {
     type Elem;
     type Error;
```
tisonkun, line 24: I suspect whether or not we need an `Error` type parameter here.

`futures::Stream` has one type parameter `Item` and the `poll_next` return type is `Poll<Option<Item>>` where `Item` can be `Result<...>`.

I can see that we fix the batch as a `Vec` which make thing is bit inflexible, but from the implementation it's always `Ok`? If you introduce an Error here, a user must decide or even define a dedicated `Error` type for using such batch stream, even if it's unnecessary.

tisonkun (commented):
I'm looking into `futures` crate occasionally, too. Comments inline.

src/futures/src/stream/batch/batch_size.rs
tisonkun: What about `WithBatchSize` and 

```rust
let stream = stream.with_batch_size(10);
stream.next();
stream.collect();
```

src/futures/src/stream/batch/result_stream.rs
tisonkun: What about

```rust
pub trait BatchResultStream {
    type Item;

    /// Returns the next `batch_size` elements.
    fn poll_next(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
        batch_size: usize,
    ) -> Poll<Vec<Self::Item>>;

    ...
}
```

... where `Item` can be `Result<..>` and it's out-of-the-box to [convert a `Vec<Result<T, E>>` to `Result<Vec<T>, E>`](https://stackoverflow.com/questions/63798662/how-do-i-convert-a-vecresultt-e-to-resultvect-e).

src/futures/src/stream/batch/batch_size.rs
huachaohuang: `with_batch_size` is a bit long for me, how about `Batched`? Like [Copied](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.copied).

src/futures/src/stream/batch/batch_size.rs
huachaohuang: I think referring to an internal field in comment is not a good idea.

src/futures/src/stream/batch/result_stream.rs
huachaohuang: I actually considered all the ideas you suggest here. But it is more straightforward for the existing use cases like `list_xxx` to return a `Result<Vec<T>, E>` than a `Vec<Result<T, E>>`. It's OK to add a more general batch `Stream` when necessary, but there are no such use cases now so I don't bother to do that yet.

src/futures/src/stream/batch/batch_size.rs
tisonkun: @huachaohuang how do you document such method then? I'm thinking of adding a pub method returning the batch_size and referring to the method, perhaps named as `batch_size()`.

src/futures/src/stream/batch/batch_size.rs
tisonkun: LGTM.

src/futures/src/stream/batch/result_stream.rs
tisonkun: You're right this is more like an internal crate than general purpose futures utility.

src/futures/src/stream/batch/batch_size.rs
huachaohuang: I think we can add comments to the type instead of explaining every method that returns batch size elements at a time.

tisonkun (approved):
LGTM

Issue #263, opened 2022-01-07
Support Shared Storage


zojw:
This is a tracking issue for implementing the cache storage

the design can be found in #247 

- [x] Prototype & Basic Trait #252
- [x] Implement Storage trait for S3 #271
- [x] #322
- [x] #323
- [x] #324
- [x] #332
- [ ] Port-back file system local storage for cache node service
- [ ] Replace mocked orchestrator by k8s backend implement

zojw:
I'd like to take 2nd item: "implement Storage trait for S3" first and investigate 3rd item after finish it

huachaohuang:
Thanks. I had some discussion with @zojw and he would like to take over the design and implementation of the cached storage once #252 gets merged.

zojw:
Hi, I'd update the issue sub-task list with more component implementation details

and I'd like to take 3rd item "Define Communication API between components and implement as Remote Storage Trait" first, it will demonstrate how components work together and final use API (trait)(for implements detail it's planning to use memory or mock implement first in this work item)

tisonkun:
@zojw could you create issues for all these subtasks? The reason as in https://github.com/engula/engula/issues/221#issuecomment-1013798397. Of course, you can always make your progress but describing the status helps participants to extend your strength.

For example, currently the S3 implementation isn't well tested and you can create an issue for it.

zojw:
The shared storage development work has been migrated to another repo as one of engula's external implements, the related issue has be replaced with https://github.com/zojw/shared-storage/issues/4, so close this first

PR #271, opened 2022-01-08
storage: implement storage trait for s3

zojw:
ref #263 

implement new Storage & futures trait for s3 storage.


src/storage/src/external/s3/storage.rs
```diff
@@ -0,0 +1,696 @@
...
+        let this = self.get_mut();
+        this.write_buf.extend(buf);
+        Poll::Ready(Ok(buf.len()))
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<io::Result<()>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
```
zojw, line 550: it's not a suboptimal solution, it's better lazy createUpload in poll_write and pre-upload when continue poll_write meet UPLOAD_PART_SIZE before close.

but it will be more complex in implement, we can do it later after other part of cache storage done

tisonkun: "It's a suboptimal ..."? Also you can capitalize the first word in a sentence :)

Cargo.lock
```diff
@@ -34,24 +34,347 @@ version = "1.0.1"
...
+[[package]]
+name = "chrono"
+version = "0.4.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "670ad68c9088c2a963aaa298cb369688cf3f9465ce5e2d4ca10e6e0098a1ce73"
+dependencies = [
+ "libc",
+ "num-integer",
+ "num-traits",
+ "time 0.1.43",
```
tisonkun, line 340: I think you should ignore `RUSTSEC-2020-0071` in `.cargo/audit.toml` for this version due to https://github.com/chronotope/chrono/pull/578 seems lack of reviews and progress.

src/storage/src/external/s3/storage.rs
zojw: > "It's a suboptimal ..."? Also you can capitalize the first word in a sentence :)

yes- - typo fixed

Cargo.lock
tisonkun: Or you may replace chrono 0.4 with time 0.3 or something (ref https://github.com/informalsystems/tendermint-rs/pull/1030).

huachaohuang:
Sent a PR https://github.com/zojw/engula/pull/4

huachaohuang (approved):
LGTM. I get some better ideas to implement this, but let's move forward for now.

src/storage/src/external/s3/storage.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    fn object_root(&self, bucket: impl Into<String>) -> String {
+        format!("{}/{}/{}", self.tenant_id, self.category, bucket.into())
+    }
+
+    // {tenant}/buckets/{bucket_name}/{object_name}
+    fn object_key(&self, engula_bucket: impl Into<String>, object: impl Into<String>) -> String {
+        format!("{}/{}", self.object_root(engula_bucket), object.into())
+    }
```
tisonkun, line 120: @zojw here you make several essential assumption. Please clarify and document it.

It seems that you're going to let the user of S3Storage use only one S3 bucket and separate logical tenant / bucket by different object key. Is this "one S3 bucket" design desirable or we may later change it? cc @huachaohuang

zojw: @tisonkun ok, I'll add the document later when next refine.

> Is this "one S3 bucket" design desirable or we may later change it?

Yes, it's desirable,  "one S3 bucket" has some discussion in https://github.com/engula/engula/discussions/256 (sorry, I should indicate it in PR description or issue), it's a bad idea to use "one S3 bucket per engula bucket" due to bucket count limit in AWS S3, the "one bucket" is enough for single usage, but for other storage vendors, this shouldn't be always true.

At last, for tenants, it should be common requirements to storage-service deployers just have one AWS account but want to support multiple engine-service clusters, and both them want to use the same bucket name, so it's reserved it in path design

PR #273, opened 2022-01-09
kernel: execute some update operations

huachaohuang:
`Kernel` should create/delete the corresponding streams/buckets on receiving updates. The atomicity is not guaranteed so far.

### DIFF
--- a/src/kernel/src/local/update_writer.rs
+++ b/src/kernel/src/local/update_writer.rs
@@ -17,24 +17,58 @@ use std::sync::{
     Arc,
 };
 
+use engula_journal::Journal;
+use engula_storage::Storage;
 use tokio::sync::broadcast::Sender;
 
 use crate::{async_trait, Error, KernelUpdate, Result, Sequence, UpdateEvent};
 
-pub struct UpdateWriter {
+pub struct UpdateWriter<J, S> {
+    journal: Arc<J>,
+    storage: Arc<S>,
     sequence: Arc<AtomicU64>,
     tx: Sender<UpdateEvent>,
 }
 
-impl UpdateWriter {
-    pub fn new(sequence: Arc<AtomicU64>, tx: Sender<UpdateEvent>) -> Self {
-        Self { sequence, tx }
+impl<J, S> UpdateWriter<J, S>
+where
+    J: Journal,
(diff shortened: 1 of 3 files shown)


tisonkun (changes requested):
I'd prefer make use of the existing `UpdateReader` instead of pollute `UpdateWriter` interface.

See also https://github.com/huachaohuang/engula/pull/4 while async rust is hard to be scoped :-(

src/kernel/src/local/update_writer.rs
```diff
@@ -17,24 +17,58 @@ use std::sync::{
...
+        for stream in &update.remove_streams {
+            self.journal.delete_stream(stream).await?;
+        }
+        for bucket in &update.add_buckets {
+            self.storage.create_bucket(bucket).await?;
+        }
+        for bucket in &update.remove_buckets {
+            self.storage.delete_bucket(bucket).await?;
+        }
         let sequence = self.sequence.fetch_add(1, Ordering::SeqCst) + 1;
```
tisonkun, line 72: Will this event still go through the channel so that the engine receive them? Why does the engine want to receive these events if so?

What if the kernel expose method just create/delete stream/bucket?

The responsibility of UpdateWriter is unclear if it severs blocking create/delete operations.

huachaohuang:
Oh, these things really don't work that way. It requires a deep understanding of RocksDB internals (and database architecture, perhaps) to understand why `Kernel` exists and how it should work. I don't think I can explain it in simple terms. The best way for me is to show it in the luna engine. And the best way for others is to learn more RocksDB internals.

tisonkun:
@huachaohuang OK. Go ahead if you're confident is the right direction.

tisonkun:
@huachaohuang I notice that the `Kernel` can still provide create/delete stream/bucket methods and send the update to the channel internally. To avoid users barely send such an update via `UpdateWriter`, the `KernelUpdate` message should be sealed when exposed.

Anyway, it's an internal optimization. We can revisit the implementation with a basically usable luna-engine or I'll dig into it later.

w41ter-l (approved):

tisonkun (approved):
No blocker.

PR #280, opened 2022-01-10
rfc: add leader-based journal

w41ter-l:
The rendered version: [leader based journal](https://github.com/w41ter-l/engula/blob/rfc_single_write_journal/docs/rfcs/20220110-leader-based-journal.md)


w41ter-l:
cc @huachaohuang

zojw:
Confirm another question~ (storage seems also have the same question

Now, we have 2 engine nodes (L1 and F2), and we add 2 engine Nodes(F3, F4) to the cluster, how do `F3, F4` know whether they need to replicate L1's data?

Or where is suitable to maintain metadata like `default-tablet-group` -> `Leader L1, Follower F2, F4, Nothing F3`?

Orchestrator could give a full list of L1, F2, F3, F4, but it doesn't know F3 no need to replicate...

w41ter-l:
@zojw Now we only need to implement single writer journal, it means that only one engine is writer, the others are readers. But once partition is added to engine,  we should add a new node to manage the destribution of partitions.

pengfeiwang-cn:
why do we a Single Write Journal?

tisonkun:
@w41ter-l @huachaohuang I noticed that we're implementing the API designed here. If the RFC is accepted, please update the status and pull request links, start a tracking issue for the work (added in the RFC) and merge this PR.

Otherwise, the RFC process is fake.

docs/rfcs/20220110-single-writer-journal.md
```diff
@@ -0,0 +1,71 @@
+# Single write journal
+
+- Status: draft
+- Pull Request:
```
tisonkun, line 4:
```suggestion
- Status: accepted
- Pull Request: https://github.com/engula/engula/pull/280
- Tracking Issue: <please create one of this>
```

tisonkun: co https://engula.zulipchat.com/#narrow/stream/306467-general/topic/random/near/267827815

cc @huachaohuang @w41ter-l

docs/rfcs/20220110-single-writer-journal.md
```diff
@@ -0,0 +1,72 @@
...
+
+A event will be replicated to all journal servers of a replication group eventually. Once an event is replicated to enough journal servers, which is specified by the replication policy, the event could be committed and applied to engine.
+
+#### Reconfiguration
+
+In general, a replication policy allows a journal server downtime unexpectedly, but does not affects the writing operations. In order to keep the availability in this situation, master would enforce leader to seal previous events and allocate new epoch so that it could change the configuration such as replication group to remove the faulted nodes.
+
+#### Follower read
+
+A leader will broadcast the committed sequence of events to all journal server, and those events is visible for reading. But here exists a gap between an event become committed in leader and an event is readable in a journal server. So a follower want to read events with consistency, it should ask the latest committed sequence from leader and wait until it receive those events.
```
tisonkun, line 62: Who is expected to perform follower read?

If it happens only during recovery, duplicate followers seems overkill.

w41ter-l: An engine follower must read the stream for the duration of its life to keep track of the leader's state. And no follower read is performed during recovery, because the recovery may execute in leader,

tisonkun: Thanks for your explanation :)

huachaohuang:
I think this RFC proposes two things: a new `SingleWriteJournal` API and an implementation of the API. I suggest that we split the RFC into two, one RFC describes the semantic of `SingleWriteJournal` (I prefer `LeaderBasedJournal`, though), another RFC describes an implementation of the new API.

docs/rfcs/20220110-single-writer-journal.md
```diff
@@ -0,0 +1,72 @@
...
+
+This has discussed in [#260](https://github.com/engula/engula/discussions/260). A trait named `SingleWriteJournal` is introduced to observe the role changes:
+
+```rust
+pub enum RoleState {
+    Leader,
+    Follower,
+}
+
+pub trait SingleWriteJournal : Journal {
```
tisonkun, line 23: Here we have two focuses:

1. For persisting data into journal server cluster, the proposal is using a quorum-based algorithm as its replication policy.
2. For supporting high availability via standby engine (journal client), the proposal is electing a leader for all engines.

If so, I suggest that we separate these two focuses and resolve them one after the other.

1. `QuorumBasedJournal`: a client side implementation that dealing with persisting data into journal server cluster.
2. ...
    * `LunaEngine` (or its internal JournalWriter, concretely) has a leader election mechanism for determining which instance is the exclusive writer of current shard, with a monotonically increasing token. When a new leader elected, it writes its token to the journal server cluster (quorum).
    * `FencedJournal`: a server side journal implementation that accepted monotonically increasing token. It only serves requests configured with the current token and accepts monotonically increasing only token updates.

w41ter-l:
I have divided this RPC into two parts. This part mainly focuses on the design of the leader based journal, the design of the shared journal will be submitted with a new RPC later.

cc @huachaohuang @tisonkun @zojw

docs/rfcs/20220110-single-writer-journal.md
huachaohuang: This seems a bit over design, at least for use cases similar to the luna engine. The semantic of `LeaderBasedJournal` is easier to use and the previous design in this RFC naturally maps to the API. It is OK to decouple the internal implementation to some reusable components like `QuorumBasedJournal` + `FencedJournal`, but that depends on the implementation.

docs/images/single-write-journal-architecture.svg
```diff
@@ -0,0 +1 @@
...rotate(0)"><tspan id="SvgjsTspan1126" dy="16" x="250.33823152218167"><tspan id="SvgjsTspan1127" style="text-decoration:;">Write</tspan></tspan></text></g><g id="SvgjsG1128"><path id="SvgjsPath1129" d="M375.99998474121094 243.0039825439453L375.99998474121094 259.8199825286865L174.79998474121095 259.8199825286865" stroke="#323232" stroke-width="1" fill="none" marker-end="url(#SvgjsMarker1130)"></path></g><g id="SvgjsG1132"><path id="SvgjsPath1133" d="M375.99998474121094 282.0039825439453L375.99998474121094 259.8199825286865L174.79998474121095 259.8199825286865" stroke="#323232" stroke-width="1" fill="none" marker-end="url(#SvgjsMarker1134)"></path><rect id="SvgjsRect1136" width="27" height="16" x="272.9919847488403" y="251.81998252868652" fill="#ffffff"></rect><text id="SvgjsText1137" font-family="微软雅黑" text-anchor="middle" font-size="13px" width="27px" fill="#323232" font-weight="400" align="top" lineHeight="16px" anchor="middle" family="微软雅黑" size="13px" weight="400" font-style="" opacity="1" y="250.19498252868652" transform="rotate(0)"><tspan id="SvgjsTspan1138" dy="16" x="286.4919847488403"><tspan id="SvgjsTspan1139" style="text-decoration:;">Read</tspan></tspan></text></g></svg>
```
huachaohuang, line 1: This file is unused in this document.

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,50 @@
...
+```rust
+pub enum RoleState {
+    Leader,
+    Follower,
+}
+
+pub trait LeaderBasedJournal : Journal {
+    type Role;
+    type Peer;
+    type StateStream: Stream<Item = RoleState>;
```
huachaohuang, line 34: I think we need some API like `wait_next` here so that the caller can wait for the next event?

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,50 @@
...
+
+    fn state(&self, name: &str) -> (Self::Role, Option<Self::Peer>);
+
+    async fn observe_state(&self, name: &str) -> Self::StateStream;
```
huachaohuang, line 38: Maybe `watch_state` is better?

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,50 @@
...
+Here is the API design of leader-based journal.
+
+```rust
+pub enum RoleState {
+    Leader,
+    Follower,
+}
+
+pub trait LeaderBasedJournal : Journal {
+    type Role;
```
huachaohuang, line 32: If these types depend on the implementation, how does a user use them if they don't know the concrete types?

docs/rfcs/20220110-leader-based-journal.md
tisonkun: As long as it's a stream of state/event, an extension like [`next()`](https://docs.rs/tokio/0.2.19/tokio/stream/trait.StreamExt.html) can achieve this out-of-the-box.

@w41ter-l Is "state" just "event" or something different?

w41ter-l: @tisonkun I've added some details, maybe able to answer your question.

docs/rfcs/20220110-leader-based-journal.md
w41ter-l: There doesn't seem to be a difference between the two, is there something I haven't noticed?

docs/rfcs/20220110-leader-based-journal.md
w41ter-l: @huachaohuang It seems `observe_state` is enough to wait for the next event, are there details that I haven't considered?

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,58 @@
...
+    fn role(&self) -> Role;
+
+    // The leader of the associated stream.
+    fn leader(&self) -> Option<String>;
+}
+
+pub trait LeaderBasedJournal : Journal {
+    type StateStream: Stream<Item = Box<dyn EpochState>>;
+
+    fn state(&self, name: &str) -> Result<Box<dyn EpochState>>;
```
tisonkun, line 44:
```suggestion
    /// Get the current state of the stream with `stream_name`.
    fn state(&self, stream_name: &str) -> Result<Box<dyn EpochState>>;
```

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,58 @@
...
+
+    async fn observe_state(&self, name: &str) -> Result<Self::StateStream>;
```
tisonkun, line 46:
```suggestion
    /// Subscribe state updates of the stream with `stream_name`.
    async fn observe_state(&self, stream_name: &str) -> Result<Self::StateStream>;
```

tisonkun: If the function return the state stream handle only, it can be a sync fn?

tisonkun (dismissed):
LGTM. Nits inline.

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,58 @@
...
+
+When the engine calls `LeaderBasedJournal::observe_state`, the `LeaderBasedJournal` starts monitoring and subscribing to the electing state transition. It will yield a `Stream` that will be fired whenever one of the electing states changes. We utilize the epoch to track state changes. Time is divided into epochs of arbitrary length, the `LeaderBasedJournal` must ensure that each epoch has only one leader.
+
+When a leader engine crashes, another machine's `LeaderBasedJournal` instance is elected as the new leader and begins to recover, eventually providing service.
+
+We can't ensure that the state returned by the `observe_state` or `state` methods is always fresh in a distributed system, but any write operations will identify this circumstance. As a result, every decision made before submitting should trigger any write operations to check for freshness.
```
tisonkun, line 58: You can discuss the semantic of APIs in sections instead of paragraphs, which separates focuses better.

Also, for significant procedure, such as read/append events, leader state changes, and recovery, you'd better include pseudo code of the procedure or a diagram. For example, in #287 I can see a new concept `Phase` that participates the recovery and leader state changes. I guess it's part of significants of this RFC?

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,58 @@
+# Leader Based Journal
+
+- Status: accepted
+- Discussion: https://github.com/engula/engula/discussions/260
+- Pull Request: https://github.com/engula/engula/pull/280
+
+## Summary
+
+In this RPC, we present a trait `LeaderBasedJournal`, which divides the users of `Journal` into two roles: a leader who could write, and followers, who only have read permission. In the same time, this trait provides a means of observing role transition.
```
tisonkun, line 9: According to the following API, role binds to a stream of the journal. Do you intend to elect leader per stream (a.k.a., in stream granularity)?

Here you write leader and followers bind to a journal, though.

docs/rfcs/20220110-leader-based-journal.md
zojw, line 46: how do we handle connection reconnection after a stream is returned?

is the caller's duty to retry or retry logic is done inside the returned stream?

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,58 @@
...
+}
+```
+
+The `LeaderJournal` doesn't affects the semantics of `Journal`, so `Journal::open_stream_writer` could be called whenever a stream isn't a leader. Of course, the implementation should guarantee that calls `StreamWriter::append` or other modifying operations will got a `Error::NotLeader`, if it isn't the stream leader.
+
+The `LeaderBasedJournal` will forwards the electing progress automatically, which the engine won't have to recognize it. However, the engine must initiate that automatic progress manually, because a journal might contains multiple streams, which could exceeds the hardware limitation if we monitors all stream's electing progress. As a result, just streams that the engine is interested in will be watched.
+
+When the engine calls `LeaderBasedJournal::observe_state`, the `LeaderBasedJournal` starts monitoring and subscribing to the electing state transition. It will yield a `Stream` that will be fired whenever one of the electing states changes. We utilize the epoch to track state changes. Time is divided into epochs of arbitrary length, the `LeaderBasedJournal` must ensure that each epoch has only one leader.
```
zojw, line 54: So, is the engines called `observe_state("stream1")` have a chance to be a follower or leader of stream1~? so for another engine that does not interest those stream1 can be avoid be elect as a follower

docs/rfcs/20220110-leader-based-journal.md
```diff
@@ -0,0 +1,58 @@
...
+}
+
+pub trait EpochState {
+    fn epoch(&self) -> u64;
+
+    // The role of associated stream.
+    fn role(&self) -> Role;
+
+    // The leader of the associated stream.
+    fn leader(&self) -> Option<String>;
```
zojw, line 38: what's the content of `String`, is it need to be logic server-id or something else~?

docs/rfcs/20220110-leader-based-journal.md
w41ter-l: retry logic should be done in inside.

docs/rfcs/20220110-leader-based-journal.md
w41ter-l: Thinks for you advices, I have separated paragraphs into sections.

I think the new concept is the implementation details, so I will introduce it in a follow up RFC.

docs/rfcs/20220110-leader-based-journal.md
w41ter-l: Thinks, I will fix it.

docs/rfcs/20220110-leader-based-journal.md
w41ter-l: Yes, only called `observe_state("stream")` would be a follower or leader of stream.

docs/rfcs/20220110-leader-based-journal.md
w41ter-l: I have no idea about which is better to fill, so I just left it a `String`. it can be changed once we found and defines a properly structs.

huachaohuang:
Closed in favor of #361. Please help to improve the stream store design in #361.

PR #282, opened 2022-01-10
journal: refactor event stream

tisonkun:
* `seek` and `truncate` should handle invalid argument:
  * otherwise, `seek` to a truncated sequence and read will return from the current very beginning event, it's not the one on the seek sequence.
  * otherwise, `truncate` a future sequence is noop but you can still append to arrive that sequence so that `truncate`'s semantic doesn't hold.
* `seek` becomes zero-based (seek_to_inclusive) so that we don't have to `+ 1` everywhere.
* `wait_next` is reimplemented, I hope the code more sane.

cc @huachaohuang @w41ter-l

### DIFF
--- a/src/journal/src/local/mem/mod.rs
+++ b/src/journal/src/local/mem/mod.rs
@@ -36,14 +36,17 @@ mod tests {
         let event1 = vec![1];
         let event2 = vec![2];
         let mut writer = j.new_stream_writer(stream_name).await?;
-        assert_eq!(writer.append(event1.clone()).await?, 1);
-        assert_eq!(writer.append(event2.clone()).await?, 2);
+
+        let expected_event1_seq = 0;
+        let expected_event2_seq = 1;
+        assert_eq!(writer.append(event1.clone()).await?, expected_event1_seq);
+        assert_eq!(writer.append(event2.clone()).await?, expected_event2_seq);
 
         let mut reader = j.new_stream_reader(stream_name).await?;
-        reader.seek(0).await?;
+        reader.seek(expected_event1_seq).await?;
         assert_eq!(reader.try_next().await?.as_ref(), Some(&event1));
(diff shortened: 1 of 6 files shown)


w41ter-l (dismissed):
nice job

src/journal/src/local/mem/mod.rs
```diff
@@ -86,4 +86,91 @@ mod tests {
...
+        let mut writer = j.new_stream_writer(stream_name).await?;
+        assert_eq!(writer.append(event1.clone()).await?, 1);
+        assert_eq!(writer.append(event2.clone()).await?, 2);
+        assert_eq!(reader.try_next().await?.as_ref(), Some(&event2));
+
+        // normal
+        reader.seek(0).await?;
+        assert_eq!(reader.try_next().await?.as_ref(), Some(&event1));
+        assert_eq!(reader.try_next().await?.as_ref(), Some(&event2));
+        reader.seek(1).await?;
```
huachaohuang, line 113: I think something is wrong here. If `writer.append(event1.clone()).await?` returns sequence 1, then we should return `event1` after seeking to sequence 1.

tisonkun: Thanks for your explanation! I just treated the return value of `append` as the total number of event appended 😂.

If the sequence return by `append` indicated the sequence number of the event just append, I can modify the implementation to fix this off-by-one error.

Updating...

tisonkun: @huachaohuang I pushed a follow-up commit to return zero-based sequence when appended. The semantic of the return value of `append` is the sequence number of the event just appended. We don't have to start by 1. Zero-based can works.

huachaohuang (approved):
Thanks, we should specify the semantics of these APIs more explicitly later.

PR #283, opened 2022-01-12
journal: add shared journal API

w41ter-l:
A defination of #280


huachaohuang (approved):
Go ahead. I will re-check these commits when more details are added.

PR #285, opened 2022-01-12
journal: add basic framework and structures for shared journal

w41ter-l:
This PR is successor of the former PR #283 .

### DIFF
--- a/src/journal/src/remote/shared/orchestrator.rs
+++ b/src/journal/src/remote/shared/orchestrator.rs
@@ -1,0 +1,30 @@
+// Copyright 2022 The Engula Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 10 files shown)


src/journal/src/remote/shared/orchestrator.rs
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use async_trait::async_trait;
+use engula_futures::stream::batch::ResultStream;
+
+use crate::{Error, Result};
+
+#[async_trait]
+pub trait Orchestrator: Send + Sync + 'static {
```
huachaohuang, line 21: Let's remove these trait bounds if they are not required to the interface. You can specify them in the specific implementation that requires them.

src/journal/src/remote/shared/orchestrator.rs
```diff
@@ -0,0 +1,30 @@
...
+    type Instance;
+    type InstanceLister: ResultStream<Elem = Self::Instance, Error = Error> + Unpin;
```
huachaohuang, line 23: Ditto

huachaohuang (approved):

PR #287, opened 2022-01-13
journal: add master defination of shared journal

w41ter-l:


w41ter-l:
Closes this in favor of #312.

PR #291, opened 2022-01-17
rfc: add shared journal

w41ter-l:


w41ter-l:
@tisonkun Shall I submit this as a design document instead of an RFC?

tisonkun:
@w41ter-l we almost exchangeably use the term "RFC" and "design document", what's the different?

If you mean items under "docs/" and items under "docs/rfcs", I think you can submit the doc under the "rfcs", and when it becomes stable as fundamental design, duplicate one into "docs/".

docs/rfcs/20220114-shared-journal.md
```diff
@@ -0,0 +1,116 @@
...
+
+##### Recovery
+
+// TODO(w41ter) This process is a bit complicated to explain, and I will add it when I have more time.
+
+#### Replacing
+
+In server side, the master will detects the fault servers.
+
+For the sealed segments, if one of it's replica is lost by fault, the master will choose a new server and via a place RPC to enforce that server to restore the replica, by reading from the health replicas. If the number of surviving replicas doesn't satisfy the replication policy's requirements, manually repair is required. Once the journal server has restored the replica, the master could atomically replace the failed replica.
```
huachaohuang, line 92: It may not be necessary to do replica recovery if we archive sealed segments to external object storage. This avoids copying data among journal servers.

w41ter-l: Yes, if the segments have already archived or could be archived, replica recovery is unnessary. And this paragraph is for segments haven't been archived.

huachaohuang: I mean, we can archive automatically when some replicas are down.

w41ter-l: This strategy is easy to achieve once we support a mechanism that archive segment to external object.

huachaohuang:
Closed in favor of #361. Please help to improve the stream store design in #361.

PR #296, opened 2022-01-18
hash: remove it since we don't have a long-term plan for it

### LINKED ISSUE #295: Remove the hash engine
The hash engine was mostly created for demonstration. Without a long-term development plan, it is not useful to keep it in the codebase. People who want to try it out can use the release branch instead.

huachaohuang:
Closes #295


tisonkun (approved):
Thanks for making the decision!

Issue #322, opened 2022-01-19
Implement gRPC Storage


zojw:
ref #263 

make both client and server sides usable again

## Phase 1:

support 1 to 1 invoke

  - [ ] Port gRPC Storage back, Single Remote gRPC Storage, adapt engula-futures  #331
    - add protobuf (like list_bucket, list_object and additional  update param)
    - implement new storage IO trait for both side 

## Phase 2:

combinate multiple phase1 result, to support invoke multiple nodes or retry between multiple nodes

  - [ ] Implement wrapper storage that can sequentially invoke multiple gRPCClient Storage
  - [ ] Implement wrapper storage that can invoke one gRPCClient Storage and retry another if fail

zojw:
The shared storage development work has been migrated to another repo, the related issue has be replaced with https://github.com/zojw/shared-storage/issues/4, so close this first

PR #331, opened 2022-01-20
storage: fix gRPC Storage

zojw:
ref #322

Make gRPC storage works again, it's a precondition for follow-up works in #322 and #263.

It should have memory and network performance questions, but we can optimize and make better trait implement after all components can work together.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -457,6 +457,9 @@ dependencies = [
  "thiserror",
  "time",
  "tokio",
+ "tokio-stream",
+ "tokio-util",
+ "tonic",
  "tonic-build",
 ]
 
--- a/src/futures/Cargo.toml
+++ b/src/futures/Cargo.toml
@@ -9,6 +9,7 @@ description = "Asynchronous utilities for Engula."
 
 [dependencies]
 futures = "0.3.19"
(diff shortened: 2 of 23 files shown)


huachaohuang:
Can you explain the necessity of `ReadFromPosExt`?

zojw:
> Can you explain the necessity of `ReadFromPosExt`?

@huachaohuang 

It wants to handle "ReadAt then return stream" situations like https://github.com/engula/engula/pull/331/files#diff-af8ffd5a16a7551820f2424415260ac0b6d522de73cc34f175f6a038fe5f124bR117

At first, gRPC only supports the standard stream that's the reason why we need to implement position-read that return stream,  the semantic should be "start at pos then read until stream close"(or limit hint), the semantic is very suitable for an AsyncRead that start from pos then maintain pos to forward read.

To implement RandomRead -> Stream transform, it's also natural to reuse existing [ReaderStream](https://docs.rs/tokio-util/0.4.0/tokio_util/io/struct.ReaderStream.html)

huachaohuang:
I think it is more complicated than necessary. For random reads, we can simply send an `offset + size` to the server and the server can simply returns a buffer instead of a stream.

zojw:
closes this, because we don't need gRPC storage after #353 merged

PR #341, opened 2022-01-21
journal: add shared journal mem server

### LINKED ISSUE #327: Add shared journal server memory implementation

w41ter-l:
This PR closes #327 .

A very rough implementation of mem journal server to facilitate client development.

### DIFF
--- a/src/journal/src/remote/shared/proto/server.proto
+++ b/src/journal/src/remote/shared/proto/server.proto
@@ -17,7 +17,8 @@ syntax = "proto3";
 package engula.journal.v1.shared.server;
 
 service SharedJournal {
-    // Store some continuous entries to a segment.
+    // Store some continuous entries to a segment. This function will create
+    // a new segment replica if it not exists.
     rpc Store(StoreRequest) returns (StoreResponse);
 
     // Read some entries from a segment.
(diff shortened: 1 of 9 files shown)


src/journal/src/remote/shared/proto/server.proto
```diff
@@ -17,7 +17,8 @@ syntax = "proto3";
 package engula.journal.v1.shared.server;
 
 service SharedJournal {
```
tisonkun, line 19: It seems that `SharedJournal` service doesn't serve the interface of `Journal` trait. What's your plan on interoperate or merging these two abstractions?

tisonkun: In another word, is a `SharedJournalServer` also a `JournalServer`? If not, what's your design of the relationship between these two concepts?

w41ter-l: In `SharedJournal`, the server is only responsible for saving `Segment`, that is, part of the function of `Stream`, and does not need to implement `Journal` trait.

It's inappropriate to call `SharedJournal` or `JournalServer` here. It behaves more like Segment Service, but I haven't figured out what to name it yet. If you have better suggestions, please feel free to issue them.

tisonkun: Then, where is the `SharedJournal` server runs on? Colocated with Kernel server? Colocated with Journal server? Or a separated node? A deployment graph here or in #291 should help.

w41ter-l: The role of the server in the code here is the journal server in the design draft #291 . However, the journal server name is ambiguous and will be confused with the grpc version of the journal server, but it has no related interface to the `Journal` trait.

PR #348, opened 2022-01-22
replace git dep to crates dep

tisonkun:
so that we don't have to clone the whole git repo for dependencies.

cc @zojw


tisonkun:
cc @huachaohuang 

also this PR fixes a CI bug that rust-toolchain isn't pinned by `rust-toolchain.toml` file.

huachaohuang (approved):
Thanks!

PR #353, opened 2022-01-26
code structure

### LINKED ISSUE #352: Remove gRPC implementation of storage and journal
BTW, for share storage and share journal implementation, as previous discussion about separate modules, I propose those implementation leaves in a separated crates and rely on the basic implementation provided by `engula-storage` or `engula-journal`.

Also, current code structure can be refactor a bit so that the root `Cargo.toml` only contains `workspace` section and we move the `engula` crate as `src/engula`. In this way, all crates under this repository are peers, one is not surprised by that one crate has a special layout. This is a code structure refactor thought, anything about publishing won't be affected.

tisonkun:
This closes #352 .

cc @huachaohuang 

If we hold all workspace members under `src/`, I don't think it's necessary to name them "engula-xxx".

@w41ter-l @zojw I don't touch the cached storage or shared journal implementation, it's your responsibility to decide where it should lay out. Feel free to comment below for discussion about code structure details.

One potential risk is that I'm unsure whether this change affects our [release procedure](https://github.com/engula/engula/blob/d335f0aa85ac38611c1d5b00b8f6b7ae9d39c9b7/docs/dev/release-guide.md). But we can update the doc if it encounters troubles.


huachaohuang (approved):
Thanks

PR #413, opened 2022-02-16
stream-engine: add Engine service interface definitions

### LINKED ISSUE #412: stream-engine: add basic client interfaces

### LINKED ISSUE #415: stream-engine: add basic master service definitions

huachaohuang:
Closes #412, closes #415


w41ter-l:
@huachaohuang Why define a proto instead of adding a rust trait like:

```
trait Engine {
     fn list_tenants();
     fn create_tenants():
     ...
}
```

huachaohuang:
After some chat with @w41ter-l , we agree to define the client interfaces instead of the service APIs first. The service APIs are actually for internal implementation, not public APIs.

huachaohuang:
Hmm, I realize that the protobuf definitions are still necessary to implement the client. So I think I can simply add the client interfaces along with this PR.

huachaohuang:
I added some basic client interfaces. If we need to add other implementations, we can mock the `MasterClient`. For example, we can simply send requests to a local master.

@w41ter-l do you think this is something we can start with?

src/stream-engine/proto/proto/master.proto
```diff
@@ -0,0 +1,125 @@
...
+  rpc Tenant(TenantRequest) returns (TenantResponse) {}
+
+  rpc Stream(StreamRequest) returns (StreamResponse) {}
+}
+
+message TenantRequest { repeated TenantRequestUnion requests = 1; }
+
+message TenantResponse { repeated TenantResponseUnion responses = 1; }
+
+message TenantRequestUnion {
```
w41ter-l, line 29: This definition seems strange to me, what is it purpose?

w41ter-l:
> @w41ter-l do you think this is something we can start with?

@huachaohuang Yes, I think it is enough.

src/stream-engine/proto/proto/master.proto
huachaohuang: It is for batch requests. Another alternative is:

```
message TenantRequest {
  oneof request {
    ...
  }
}

message BatchTenantRequest {
  repeated TenantRequest requests = 1;
}
```

huachaohuang: Hmm, I think we can drop the batch for now.

w41ter-l: What are the scenarios and benefits of batch requests?
In the implementation of underlying RPC, the overhead of batch and non-batch is only a few bytes.

huachaohuang: It is for future extension, but I think we can drop it for now.

w41ter-l (approved):

src/stream-engine/proto/proto/master.proto
huachaohuang: Hmm, another thing is that we will need to add some shared context for `TenantRequestUnion` in `TenantRequest`, like the tenant in `StreamRequest`. So I suggest we leave it as it is.

PR #458, opened 2022-02-22
stream-engine: add segment store service defination

### LINKED ISSUE #436: stream-engine: add segment store service defination

w41ter-l:
This PR closes #436.


src/stream-engine/proto/build.rs
```diff
@@ -14,5 +14,6 @@
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     tonic_build::configure().compile(&["proto/master.proto"], &["proto"])?;
+    tonic_build::configure().compile(&["proto/store.proto"], &["proto"])?;
```
huachaohuang, line 17:
```suggestion
    tonic_build::configure().compile(&["proto/master.proto", "proto/store.proto"], &["proto"])?;
```

src/stream-engine/store/src/error.rs
```diff
@@ -0,0 +1,39 @@
...
+pub enum Error {
+    #[error("{0} is not found")]
+    NotFound(String),
+    #[error("{0} already exists")]
+    AlreadyExists(String),
+    #[error("invalid {0}")]
+    InvalidRequest(String),
+}
+
+impl From<Error> for Status {
```
huachaohuang, line 28: I tried to directly use `tonic::Status` as the error type in other components. It avoids some conversions between components. Just a suggestion, though.

huachaohuang (approved):

src/stream-engine/store/src/error.rs
w41ter-l: Exposing `tonic::Status` will force users to rely on tonic, I think it would be better to do some simple conversions.

huachaohuang (approved):

PR #484, opened 2022-02-24
server: handling ctrl-c signal

### LINKED ISSUE #483: server: handling ctrl-c signal
So that a ctrl-c signal from docker frontend can kill the process.

tisonkun:
This closes #483.

On the fly it supports addrs with any hostname.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,9 +22,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.53"
+version = "1.0.55"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "94a45b455c14666b85fc40a019e8ab9eb75e3a124e05494f5397122bc9eb06e0"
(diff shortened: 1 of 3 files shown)


huachaohuang:
Doesn't the current version simply quit on ctrl-c?

tisonkun:
@huachaohuang if you run in host, ctrl-c kill the process. If you run with `docker`:

```bash
docker run --rm -it -p 21716:21716 ghcr.io/engula/engula server start
```

The process will ignore the signal.

cc @zheaoli do you have ideas why behaviors differ from envs?

huachaohuang:
I think [this](https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html#trouble-signaling-pid-1) explains the problem. Simply speaking, the fallback signal handler will not be called when we are PID 1. So we need to handle the signal on our own in this case.

huachaohuang (approved):
LGTM, handling the signal seems no harm.

Zheaoli:
I think it's good to handle the signal when we are PID 1, or we need a init process like tini or else instead.

PR #488, opened 2022-02-25
stream-engine: add replication algorithm

### LINKED ISSUE #480: stream-engine: add replication algorithm

w41ter-l:
This PR closes #480.


huachaohuang (approved):
Wow, this is a huge amount of work. Please share the detailed design later and then we can review the design and implementation together.

PR #493, opened 2022-02-26
test: add integration tests

### LINKED ISSUE #226: Add test for engula binary

tisonkun:
This closes #226.


src/client/tests/it/api.rs
```diff
@@ -0,0 +1,45 @@
...
+// limitations under the License.
+
+use anyhow::Result;
+use engula_client::Blob;
+
+use crate::create_universe;
+
+#[tokio::test]
+#[ignore]
+async fn test_apis() -> Result<()> {
```
tisonkun, line 22: Can extend to more tests.

src/engula/src/server.rs
```diff
@@ -46,33 +50,53 @@ impl SubCommand {
 
 #[derive(Parser)]
 struct StartCommand {
-    #[clap(long, default_value = "0.0.0.0:21716")]
-    addr: String,
+    #[clap(long, default_value = "21716")]
+    port: u64,
+    #[clap(long, default_value = "21715")]
+    liveness_port: u64,
```
tisonkun, line 56: As a server, we always run on localhost, thus only `port` is meaningful config. And @zheaoli suggests me that we can always bind on "0.0.0.0".

src/engula/src/server.rs
```diff
@@ -46,33 +50,53 @@ impl SubCommand {
...
+        let kernel = tonic::transport::Server::builder()
+            .add_service(health_server)
             .add_service(transactor)
             .add_service(object_engine_master)
             .add_service(stream_engine_master)
-            .serve_with_incoming(TcpListenerStream::new(listener));
+            .serve_with_incoming(kernel_stream);
 
-        info!(message = "Starting Engula server...", %addr);
+        let liveness =
```
huachaohuang, line 82: Why do you add both gRPC health service and another liveness server?

src/client/tests/it/main.rs
```diff
@@ -0,0 +1,47 @@
...
+            _ => tokio::time::sleep(Duration::from_secs(interval)).await,
+        }
+    }
+
+    Err(Error::msg(format!("Exceeds retry times: {}", retry)))
+}
+
+async fn create_universe() -> Result<Universe> {
+    let port = option_env!("ENGULA_UNIVERSE_PORT").unwrap_or("21716");
+    let uv = Universe::connect(format!("http://0.0.0.0:{}", port)).await?;
```
huachaohuang, line 45: Why not simply retry this?

src/engula/src/server.rs
huachaohuang: Users may not want to bind to all interfaces due to security issues.

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2022 The Engula Authors.
```
huachaohuang, line 1: I suggest moving these files to a sub-directory to keep the root clean.

src/client/tests/it/api.rs
```diff
@@ -0,0 +1,45 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use anyhow::Result;
+use engula_client::Blob;
+
+use crate::create_universe;
+
+#[tokio::test]
+#[ignore]
```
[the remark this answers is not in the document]
PsiACE: It looks like you've deliberately ignored all integration tests, why?

src/engula/src/server.rs
tisonkun: What do you mean by bind all interface?

Could you provide an example that bind to an address except 0.0.0.0 significantly?

src/engula/src/server.rs
tisonkun: For running a simplest it case, we don't have to implement it. Let me remove it as overkill.

src/client/tests/it/main.rs
tisonkun: For current usage it can be done in the way you describe, let me simplify it.

src/client/tests/it/api.rs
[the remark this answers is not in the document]
tisonkun: Integration tests require running services which differs from unit tests.

I actually have two ideas here. 

The one is as this and run all unit tests by `cargo test --workspace`.

The other could be not to ignore any test and running test with exact package and pattern for both unit tests and integration tests.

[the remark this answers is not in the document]
PsiACE: Got it, in fact, this is equivalent to providing control to run unit tests and integration tests separately.

I think this is great, especially when there is a relatively clear boundary between unit tests and integration tests.

src/engula/src/server.rs
huachaohuang: For example, if my node has a public IP and a private IP, using "0.0.0.0" will bind to both IPs, which means that other people can access the service in my node with the public IP. If that is not what I want, I can specify the exact private IP instead of "0.0.0.0".

tisonkun: OK. I'll revert this change.

docker-compose.yml
tisonkun: Move `docker-compose-dev.yaml` to `tools/ci` and remove this file. I think any user can easily write their own and it's only valuable to provide an official one when we have more complex logic to deliver.

src/client/tests/it/main.rs
tisonkun: Updated.

src/engula/src/server.rs
tisonkun: Updated. Although, we may want to implement similar logics later.

1. For a liveness service, it report that the instance is alive. And we'd better implement it in a way that can be easily verified handy.
2. For health check, it's more complex to define readiness for kernel services, for example, a recovered cooperator may have to rebuild its state. A health service follow gRPC standard enables we to report health status of each gRPC service and `tonic_health` provides a heath reporter which can be cloned and passed to our kernel services for reporting. However, it can be an overkill for now.

PsiACE (approved):

src/engula/src/server.rs
```diff
@@ -50,27 +52,32 @@ struct StartCommand {
...
 
-        info!(message = "Starting Engula server...", %addr);
+        info!(message = "Starting Engula server...", %kernel_addr);
 
         tokio::select! {
-            res = serve => {
+            res = kernel => {
                 if let Err(err) = res {
-                    error!(cause = %err, "Fatal error occurs!");
+                    error!(cause = %err, "Kernel: Fatal error occurs!");
```
huachaohuang, line 80: Users (including me) have no idea what is a kernel. Seems you don't need to change this file at all.

huachaohuang (commented):
Rest LGTM

src/engula/src/server.rs
tisonkun:
```suggestion
                    error!(cause = %err, "Fatal error occurs!");
```

tisonkun: @huachaohuang reverted.

huachaohuang (approved):

PR #557, opened 2022-03-04
object-engine: add stream file writer & reader in filestore

### LINKED ISSUE #556: object-engine: add manifest for lsmstore & master
Manifest will be used in both lsmstore and master to maintain metadata info for bucket and lsm-tree in them

### LSMStore Manifest

- [x] (0.4) Local disk manifest implement  #586

###  Master Manifest

- [ ] (0.4) Use an embedded LSMStore to maintain manifest info

zojw:
ref #556

Extend `bucket` to support two new stream-file operations

- `new_stream_writer`:  support append write to the file
- `new_stream_reader`: support sequence read from the appendable file

In `local` package, it use the local filesystem's "append write" and "sequence read" operations to implement those two methods

And it also creates a new `external/s3` and simulates stream-file operation on blob filesystem(no append)

In detail, it introduces another trait(https://github.com/engula/engula/pull/557#discussion_r820021241) -- `BlobStore` to the abstract non-append filesystem with two methods(`new_random_reader` and `new_sequence_writer`but only accept path as argument ), and it also provides a `BlockStreamExt` that acted as a bridge from BlobStore to stream-file, so filesystem like Azure Blob can implement two new methods by its unique append support, but filesystems like S3 can use `BlockStreamExt` to impl two stream-file methods quickly and with less duplicated code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -249,6 +249,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
+name = "fuchsia-cprng"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a06f77d526c1a601b7c4cdd98f54b5eaabffc14d5f2f0296febdc7f357c6d3ba"
+
+[[package]]
 name = "futures"
 version = "0.3.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -612,7 +618,9 @@ dependencies = [
 name = "object-engine-filestore"
 version = "0.3.0"
 dependencies = [
+ "bytes",
  "object-engine-common",
+ "tempdir",
  "tokio",
 ]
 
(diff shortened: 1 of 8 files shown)


src/object-engine/filestore/src/local/store.rs
```diff
@@ -19,22 +19,43 @@ use tokio::fs;
...
 
+#[derive(Clone, Copy)]
+pub enum DiskMode {
+    Appendable,
+    OverwriteOnly,
+}
+
+#[derive(Clone)]
+pub struct Options {
+    pub disk_mode: DiskMode,
```
huachaohuang, line 30: I think "disk mode" is not an option, it's the characteristic of the underlying storage system. This local implementation is based on a local file system. And we have no plan to support an "overwrite only" file system. If I understand it right, you are trying to build an abstraction based on other cloud storage services. If so, you should do it in other modules, instead of mixing them here.

I think we can do it with another level of abstraction. For example, we can add a blob store trait that only provides random reads and sequential writes. Then implement the file store trait for those who implement the blob store trait with the stream read/write.

huachaohuang (commented):
I suggest splitting this PR into two. The first one adds the necessary file store interfaces, and the other adds the manifest.

zojw:
we'd better impl local-only version for lsmstore first, so close it and replace it with #586

PR #575, opened 2022-03-09
stream-engine: store design doc

### LINKED ISSUE #574: stream-engine-store: design

w41ter:
This PR closes #574 .


w41ter-l:
@huachaohuang 
I have add some images to illustrate the design ideas of stream engine stores. See https://github.com/w41ter-l/engula/blob/stream-engine-store-design-doc/docs/stream-engine.md#read-and-write. 
Do you have any suggestions?

huachaohuang:
@w41ter-l I am thinking about it. A few questions so far:

- How many streams are we going to serve with one segment store?
- How do you organize the file layout of tenants/streams? Do you store each stream separately or store all streams of a tenant together?
- Do we have an upper limit of unacked events? If we only tolerant a limited unacked events, maybe we can simply hold them in memory and don't need to consider flushing them.

huachaohuang:
BTW, the Sorted Events Table (SET) looks nice :)

w41ter-l:
> How many streams are we going to serve with one segment store?

No limitation, the smallest unit which store serves is the segment, and segment would come from any stream.
This is determinated by master's placement policy, we can put segments of the same stream together as much as possible, or we can choose other balancing strategies.

> How do you organize the file layout of tenants/streams? Do you store each stream separately or store all streams of a tenant together?

For simplicity, each store serves only one tenant. The stream segments are stored together, just like a db.

> Do we have an upper limit of unacked events? If we only tolerant a limited unacked events, maybe we can simply hold them in memory and don't need to consider flushing them.

Because store will serves segments come from different streams, with different write rates, it is possible that when flushing, there are still unacked events in a slower segment. If the unacked events are kept in memory, it is necessary to consider how to recycle the log file. A simple way is to rewrite the unacked events to the log again, but this will cause unexpected write amplification. Another way (my choice) is to store it separately from acked events.

huachaohuang:
According to the upper-level design, a database (tenant) will only be served by a few cooperator groups (maybe just one). Each cooperator group only uses one stream to store logs for a database. In this case, each tenant in the stream engine will have only a few streams. And the write rate of each stream can be as large as 1 GB/s. So I think we can consider storing each stream separately, which avoids a few complexities described above.

Issue #504, opened 2022-02-28
Fix crate dependencies problem for publishing


huachaohuang:
There are two problems when releasing 0.3:

1. It turns out that we still have to publish all internal crates because the public crates depend on them
    - This can be fixed by updating all internal crates and dependencies, specifying the necessary information in their Cargo.toml files
2. The build scripts in engula-supervisor and engula-cooperator don't work when publishing, because they depend on external `.proto` files in engula-apis
    - I temporarily fixed this by copying the required `.proto` files to engula-supervisor and engula-cooperator. I am trying to figure out a better solution here.

TODO:

- [ ] #505
- [ ] Fix build scripts with external `.proto` files

PR #605, opened 2022-03-15
engula/supervisor: implement the new Engula APIs

### LINKED ISSUE #604: engula/supervisor: implement the new Engula APIs

huachaohuang:
Closes #604

### DIFF
--- a/src/engula/supervisor/engula/v1alpha
+++ b/src/engula/supervisor/engula/v1alpha
@@ -1,0 +1,1 @@
+../../apis/engula/v1alpha
--- a/src/README.md
+++ b/src/README.md
@@ -11,7 +11,11 @@ Dependencies:
 ```
 engula
   ||
-supervisor/transactor/cooperator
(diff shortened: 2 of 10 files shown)


src/engula/supervisor/engula/v1alpha
```diff
@@ -0,0 +1 @@
+../../apis/engula/v1alpha
```
tisonkun, line 1: I think we can use `paths` instead of `paths-ignore` in license checker config. It seems only rust files and yaml files so far we'd like to confirm the copyright.

Although, I'm curious why this marker alike file is necessary.

tisonkun: or you may just ignore all matches `**/v1alpha`

huachaohuang: This is a symbolic link file to avoid the second problem described in https://github.com/engula/engula/issues/504.

PR #607, opened 2022-03-15
ci: use paths instead of paths-ignore in licenserc.yml

### LINKED ISSUE #606: Consider using "paths" instead of "paths-ignore" in licenserc.yml
### Search before asking

- [X] I had searched in the [issues](https://github.com/engula/engula/issues) and found no similar issues.

### Motivation

We have a bunch of "paths-ignore" in the [licenserc.yml](https://github.com/engula/engula/blob/main/tools/ci/licenserc.yml) and we occasionally need to add more files there.

Since we mostly only care about licensing files in the `src` directory, maybe we can use "paths" instead of "paths-ignore" instead.

Ref: https://github.com/engula/engula/pull/605#discussion_r826565204

### Solution

Add `src` to "paths" and remove the "paths-ignore". Not sure if there is anything else we need to include yet.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

huachaohuang:
Closes #606

### DIFF
--- a/tools/ci/licenserc.yml
+++ b/tools/ci/licenserc.yml
@@ -16,17 +16,6 @@ header:
   license:
     spdx-id: Apache-2.0
     copyright-owner: The Engula Authors
-  paths-ignore:
-    - 'docs'
-    - '**/*.md'
-    - '.dockerignore'
-    - '.gitignore'
-    - '.gitmodules'
-    - 'Cargo.lock'
-    - 'Cargo.toml'
-    - 'LICENSE'
-    - 'rust-toolchain.toml'
-    - 'rustfmt.toml'
-    - 'taplo.toml'
-    - src/engula/cooperator/engula/v1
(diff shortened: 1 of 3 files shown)


tools/ci/licenserc.yml
```diff
@@ -16,17 +16,6 @@ header:
...
-    - 'Cargo.toml'
-    - 'LICENSE'
-    - 'rust-toolchain.toml'
-    - 'rustfmt.toml'
-    - 'taplo.toml'
-    - src/engula/cooperator/engula/v1
-    - src/engula/supervisor/engula/v1
+  paths:
+    - src/**/*.rs
+    - src/**/*.proto
```
tisonkun, line 21:
```suggestion
    - **/*.rs
    - **/*.proto
    - **/*.yml
```

I'd prefer to include yaml files also.

BTW, you don't have to make changes on other files above.

PR #648, opened 2022-04-07
bin: replace multi-thread with current-thread mode

w41ter:


tisonkun:
I'm thinking that whether we can manually desugar the macro and make the thread model more configurable.

src/bin/src/server.rs
```diff
@@ -44,7 +44,7 @@ struct StartCommand {
 }
 
 impl StartCommand {
-    #[tokio::main]
```
huachaohuang, line 47: Oh, this is a mistake. The server already handles the runtime at https://github.com/engula/engula/blob/dev/src/server/src/worker.rs#L55. We can simply remove this.

huachaohuang:
This is fixed by #651

PR #649, opened 2022-04-07
feat: Support del command

tisonkun:

cc @huachaohuang @w41ter


tisonkun:
However, `redis-benchmark` doesn't have a case on `DEL` command. It's for the milestone that we should support:

* SET
* GET
* DEL

commands.

huachaohuang (approved):
Thanks, LGTM.

PR #652, opened 2022-04-08
bench: introduce criterion bench framework

tisonkun:

This is an attempt to introduce micro benchmark to the engine crate. I'm unsure how to make a baseline of the redis part or we can just compare to ourselves.

You can run locally with:

```
cargo criterion -p engula-engine 
```

The output from my environment is:

```
Gnuplot not found, using plotters backend
db_get_group/db_get     time:   [986.73 us 1.0234 ms 1.0666 ms]                                
                        thrpt:  [9.3754 Melem/s 9.7713 Melem/s 10.135 Melem/s]
                 change:
                        time:   [-20.185% -12.372% -5.1466%] (p = 0.00 < 0.05)
                        thrpt:  [+5.4258% +14.119% +25.289%]
                        Performance has improved.


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

cc @w41ter @huachaohuang

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,16 +50,54 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bstr"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba3569f383e8f1598449f1a423e72e99569137b47740b1da11ef19af3d5c3223"
+dependencies = [
+ "lazy_static",
+ "memchr",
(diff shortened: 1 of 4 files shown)


tisonkun:
@w41ter I wonder whether we will redesign the `Db` struct and possible we can postpone the bench at that time. This can be referred to implement then.

For example, currently the `Db` is simply a wrapper of `HashMap<Bytes, Bytes>` but for supporting `List` and `Map` it can be changed.

In other word, what's your idea on functionality and performance concern should be considered for `Db` struct?

huachaohuang:
I think we can add this when we have a proper engine implementation. The current `Db` is too naive to be evaluated.

tisonkun:
@huachaohuang make sense. Closed.

PR #653, opened 2022-04-08
server: avoid blocking writes

w41ter:
Since awaiting a write requires waiting for the write to complete before doing anything else, this PR removes that restriction.

Now, writes are added to the `reply_buffer` first, and when the buffer is full it tries to write to the kernel. If the socket is busy at this time, save these buffers to `reply_buffer_queue`, and try to write when the socket is writable.

### DIFF
--- a/src/server/src/connection.rs
+++ b/src/server/src/connection.rs
@@ -12,15 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::io::{self, Cursor};
-
-use bytes::{Buf, BytesMut};
-use tokio::{
-    io::{AsyncReadExt, AsyncWriteExt, BufWriter},
-    net::TcpStream,
+use std::{
+    collections::VecDeque,
+    io::{self, Cursor, ErrorKind, IoSlice},
 };
 
+use bytes::{Buf, BytesMut};
+use smallvec::SmallVec;
+use tokio::net::TcpStream;
+
 use crate::{Frame, FrameError};
+
+const WRITE_BUFFER_CAPACITY: usize = 16 * 1024;
 
 /// Send and receive `Frame` values from a remote peer.
 ///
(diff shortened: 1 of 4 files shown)


src/server/src/connection.rs
```diff
@@ -236,17 +243,167 @@ impl Connection {
...
+                self.continue_writes().await?;
+            }
+
+            if ready.is_readable() {
+                return self.stream.try_read_buf(&mut self.buffer);
+            }
+        }
+    }
+
+    async fn continue_writes(&mut self) -> io::Result<()> {
```
tisonkun, line 354: Why do we have two write path `continue_writes` and `flush`? Also I don't find a callback registered on `Interest::WRITABLE` valid. It seems that flushing remaining bytes is always driven by a next read or write.

w41ter: 1. `Flush` is called before `WriteFrame` or waiting for read ready. It is responsible for submitting a full buffer to the kernel. If ·WouldBlock· is returned during this process, the buffer will be stored in `reply_buffer_queue`. But `continue_writes` is responsible for submitting entirely `reply_buffer_queue` to kernel as much as possible.
2. `continue_writes` is that callback.

src/server/src/connection.rs
```diff
@@ -236,17 +243,167 @@ impl Connection {
...
+            self.flush()?;
+        }
+        Ok(())
+    }
+
+    // Ensure the encoded frame is written to the socket. The calls above
+    // are to the buffered writes. Calling `flush` writes the remaining contents of the buffer to
+    // the socket.
+    fn flush(&mut self) -> io::Result<()> {
+        if !self.interest_writable {
```
zojw, line 287: Why not try a `flush_buf` after `interest_writable = true`?~

For example,  we first meet `WouldBlock`when writing req1's response, but at the same time, req2, req3....req5 is on the wire, so `try_read_buf` should return immediately and give no chance to run `continue_writes` until no pending request on the wire? but it should better to write req1, req2's response after req2 done and wire is writable?~

huachaohuang:
The implementation looks a bit ugly. We don't have to rush all the optimizations in the dev branch. You can keep evaluating things in your branch until you have a satisfactory implementation.

PR #654, opened 2022-04-12
Add io-uring support

huachaohuang:
We need to support mio so that users can use it in other platforms.


src/server/src/uio/connection.rs
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        self.num_inflights == 0
+    }
+
+    fn process(&mut self) {
+        while let Some(frame) = self.rbuf.parse_frame() {
+            let cmd = Command::from_frame(frame).unwrap();
+            let reply = cmd.apply(&self.db).unwrap();
+            self.wbuf.write_frame(&reply);
```
tisonkun, line 87: @huachaohuang we may not call `unwrap` here since we don't support many commands now. If we call `unwrap`, instead of returning client an Error response, it will panic.

tisonkun: My mistake. The panic I meet in local env isn't caused by these lines.

PR #657, opened 2022-04-15
engine: add basic object and record framework

w41ter:
Objects in memory are divided into three categories:

1. object: concept similar to redis
2. record: a contiguous piece of memory that stores value, list node, etc.
3. Ordinary memory: such as buckets and control applied by hashbrown

The object and record are aligned by 8 bytes, and the highest 16 bits in the first 8 bytes are the tag. tag records the type of memory ( object or record ), whether it is tombstone. The last 48 bits of the first 8 bytes of the record store the associated object, which is used to quickly find the associated object when the log structure allocator performs compaction.

All objects are stored in the key space, while the user-specified key is stored at the end of the object object.


src/engine/src/objects/mod.rs
```diff
@@ -0,0 +1,467 @@
...
+    /// TODO(walter)
+    unsafe fn as_mut<'a, T>(&self) -> &'a mut T {
+        &mut *(self.ptr.as_ptr() as *mut T)
+    }
+}
+
+#[derive(Clone)]
+pub struct ObjectRef {
+    raw: RawObject,
+    vtable: &'static ObjectVTable,
```
huachaohuang, line 269: Does the tag in the object meta have a one-to-one mapping with a specific `vtable`? If it does, the `vtable` is redundant. We can convert the object to a concrete type using only the tag and then do whatever we need.

src/engine/src/objects/mod.rs
```diff
@@ -0,0 +1,467 @@
...
+    pub fn clear_deadline(&mut self) {
+        self.deadline = u64::MAX;
+    }
+}
+
+#[repr(C)]
+pub struct Object<T: ObjectType> {
+    pub meta: ObjectMeta,
+    value: T,
+    key: [u8; 0],
```
huachaohuang, line 142: Why not put the key before the value? So that we can access the key without knowing the value type.

huachaohuang (commented):
Some comments so far. I will take more time to think over it.

huachaohuang:
BTW, [this](https://github.com/redis/redis/blob/unstable/src/object.c#L370) is how Redis handles deallocation. Seems there are only a few fundamental object types.

src/engine/src/objects/mod.rs
w41ter: If the key is placed in the front, then the object cannot contain the T object, and the information of T cannot be saved in the type. Then using Object<T> is inconvenient.

huachaohuang:
Let's move forward and refactor it later.

PR #665, opened 2022-04-19
feat: support connection timeout 

### LINKED ISSUE #664: Support connection timeout
We should close the connection if it is idle for a while. Refer to the `timeout` configuration in `redis.conf`.

tisonkun:
This closes #664.

We check whether a connection is timeout on each tick.

I notice some common logic between the mio impl and uio impl. Possibly @zojw you may also be interested to factor out some shared code.

cc @huachaohuang

### DIFF
--- a/src/bin/src/server.rs
+++ b/src/bin/src/server.rs
@@ -58,8 +58,11 @@ impl From<DriverMode> for engula_server::DriverMode {
 struct StartCommand {
     #[clap(long, default_value = "127.0.0.1:21716")]
     addr: String,
+
     #[clap(long, default_value = "mio", arg_enum)]
     driver_mode: DriverMode,
+
+    /// Close the connection after a client is idle for N seconds (0 to disable).
     #[clap(long, default_value = "0")]
     timeout: u64,
 }
--- a/src/bin/src/server.rs
+++ b/src/bin/src/server.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::time::Duration;
+
 use anyhow::Result;
 use clap::Parser;
 use engula_server::Config;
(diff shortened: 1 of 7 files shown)


src/bin/src/server.rs
```diff
@@ -58,13 +60,20 @@ struct StartCommand {
     addr: String,
     #[clap(long, default_value = "mio", arg_enum)]
     driver_mode: DriverMode,
+    #[clap(long, default_value = "0")]
```
tisonkun, line 63: We may add a help info to show that the config is in seconds and `0` means no timeout.

src/server/src/uio/server.rs
```diff
@@ -82,6 +85,17 @@ impl Server {
             }
         }
 
+        if let Some(timeout) = self.connection_timeout {
```
zojw, line 88: `uio` still needs some mechanism to let `io.wait(1)` (line 67) return immediately when meet the timeout condition even if there is no request(e.g. timeout=10s, but no request in line, `io.wait(1)` should stop block after 10s)

one way should push `timeoutOp`(https://github.com/tokio-rs/io-uring/blob/master/io-uring-test/src/tests/timeout.rs 

(ps: there are some a  worth reading blogs about "when to push timeoutOp trade-off"  https://www.ihcblog.com/rust-runtime-design-3/ | Time Driver 设计

huachaohuang: Yes, I think we need a time driver. Checking every connection on every tick and getting the clock frequently is not a good idea. We can see how other projects implement it.

tisonkun: That should be nice. Currently, both mio and uio implementation are single thread. If we'd like to introduce an async worker/driver and coordinate between threads, we may need a thread model design first.

huachaohuang: To handle connection timeout, I think a server cron job is good enough. So we need a mechanism to run a cron job regularly and maintain various server states in the job. You can check how Redis or Nginx implement the cron job.

huachaohuang: They usually also implement a time cache to avoid getting the time from kernel.

tisonkun: @huachaohuang after we implement a server cron job, it's normally triggered in the main event loop as this draft implementation (per tick, I mean). I agree that we can introduce an abstraction for server cron job and retrieve the current time from kernel only once. But the codepath is almost the same as current implementation. Did I get it right?

Or if we should treat (connection) timeout event and io event equally, we should introduce an abstraction for generic "event" and thus an event loop based on blocking queue or mpsc channel.

tisonkun: FYI Redis's `aeProcessEvents` processes time event every turn (per tick).

huachaohuang (approved):
All right, I don't mind getting it merged for now. But we should do it better later.

PR #667, opened 2022-04-21
engine: add log structured allocator

w41ter:
Here is a rough lsa implementation. There are some features that have not been added:
1. compaction
2. gc
3. ...

### DIFF
--- a/src/engine/src/alloc/mod.rs
+++ b/src/engine/src/alloc/mod.rs
@@ -216,7 +216,7 @@ impl Lsa {
         let mut segment_ptr = NonNull::new_unchecked(segment_addr as *mut Segment);
         let segment = segment_ptr.as_mut();
         segment.free(layout);
-        if !segment.active && Self::should_compact(&segment) {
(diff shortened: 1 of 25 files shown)


src/engine/src/alloc/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+        }
+        assert_ne!(segment_addr, 0);
+        let mut segment_ptr = NonNull::new_unchecked(segment_addr as *mut Segment);
+        let segment = segment_ptr.as_mut();
+        segment.free(layout);
+        if !segment.active && Self::should_compact(segment) {
+            self.wake();
+        }
+        if !segment.compacting {
+            self.might_reuse_segment(segment_ptr);
```
zojw, line 223: `segment.compacting==false` may no longer hold inside method, because `self.wake()` compaction thread can change it at the same time ~?

w41ter: fixed

PR #668, opened 2022-04-22
feat: config from file 

### LINKED ISSUE #663: Add a configuration file
We can call the configuration file `engula.toml` and the server can start with `engula server start -c engula.toml`.
We can also port some configurations from `redis.conf`.

tisonkun:
This closes #663.

cc @huachaohuang @w41ter @zojw 

Two design decisions are made here:

1. Configs from args or file can be present together. We prefer those from args to those from file.
2. Thus, we should detect whether a config is passed from args, so the default value of args cannot set. Otherwise we don't know it's configured by default or by users.

Code for parsing the toml file can be refactor if we grow our configs.

### DIFF
--- a/src/bin/src/server.rs
+++ b/src/bin/src/server.rs
@@ -78,6 +78,7 @@ impl StartCommand {
 
 const DEFAULT_CONNECTION_TIMEOUT: Option<Duration> = None;
 
+/// Merge configs from args and files, and prefer those from args to those from file.
 fn merge_configs(args: StartCommand, values: Option<Value>) -> Result<Config> {
     let addr = args.addr;
 
--- a/src/bin/src/server.rs
+++ b/src/bin/src/server.rs
@@ -14,9 +14,9 @@
 
 use std::{fs, time::Duration};
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use clap::Parser;
-use engula_server::Config;
+use engula_server::{Config, ConfigBuilder};
 use toml::Value;
 
 use crate::argenum::DriverMode;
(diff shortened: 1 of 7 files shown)


tisonkun:
I'm not sure whether it's the better solution for configs. So it's open to suggestions on the design decisions.

src/bin/src/server.rs
```diff
@@ -63,22 +49,64 @@ struct StartCommand {
...
+                        "timeout should be non-negative integer, but {:?}",
+                        timeout
+                    ))
+                }
+            }
+        }
+    }
+    if let Some(timeout) = args.timeout {
+        connection_timeout = Some(Duration::from_secs(timeout as u64));
+    }
```
tisonkun, line 105: It's an option that we factor out config to a new crate and write a config build for such painter config algorithm (all fields settable).

tisonkun: But it can be an overkill until we do introduce so much configs that we should factor out to a dedicated crate.

huachaohuang:
> Two design decisions are made here:
> 
> 1. Configs from args or file can be present together. We prefer those from args to those from file.
> 2. Thus, we should detect whether a config is passed from args, so the default value of args cannot set. Otherwise we don't know it's configured by default or by users.

Sounds right to me. I also think we should only export the most basic configurations as command-line arguments, and leave the advanced ones to the configuration file.

src/bin/src/server.rs
tisonkun: Anyway I introduce a builder to factor out such code.

I'll try to investigate whether we can leverage toml serde.

w41ter (approved):

src/bin/src/server.rs
tisonkun: Replace manually retrieval with toml's serde.

tisonkun:
I think this PR is ready to merge. Please take a look.

huachaohuang (approved):

PR #670, opened 2022-04-26
engine: object add lru field

w41ter:


huachaohuang:
Please resolves the conflicts.

src/engine/src/db.rs
```diff
@@ -30,22 +30,35 @@ unsafe impl Sync for Db {}
 
 #[derive(Default)]
 struct DbState {
+    lru: u32,
```
huachaohuang, line 33: What does this `lru` field mean?

w41ter: A count of logical clocks equal to the LRU value for each insertion of a new object.

PR #685, opened 2022-05-19
feat: prototype master 

tisonkun:
Basic tonic based master instance.

According to offline discussion, we should still investigate:

1. A database encapsulation of all collections consisting of metadata in master group. So far, it implements nodes collection only.
2. Support lookup group (desc) associated with the given key (range).
3. Support create new collection and alloc one or more replica groups for it. 
4. Integrate with raft algorithm for running a master group instead of single master instance.
5. Integrate with gossip algorithm for propagating metadata.

This PR also sets up basic CI workflow and you may look at https://github.com/zeke/semantic-pull-requests#configuration for what I propose for a consistent commit message.

cc @huachaohuang @w41ter @zojw


huachaohuang:
Thanks, I prefer to set up the CI workflow first, and I created a PR https://github.com/engula/engula/pull/688. For the commit message style, I think we can keep it simple for now and reconsider your proposal when necessary.

tisonkun:
@huachaohuang ok. I'll update this PR later.

tisonkun:
Closed. Wait for other patches go in first.

PR #686, opened 2022-05-20
proto: add basic api defination

w41ter:
I tried to add the basic RPC skeleton, if there is any missing method, please feel free to add it.

/cc @tisonkun @huachaohuang @zojw


w41ter:
close and in favor of #693 .

PR #687, opened 2022-05-23
docs: add top-level design

huachaohuang:
[Rendered version](https://github.com/huachaohuang/engula/blob/design/docs/design.md)


tisonkun:
@huachaohuang shall we merge this doc and keep update it during our implementation?

@w41ter I'd like to know the behavior if a node receive a group request of which the group is not on that node. So far, I don't find a clear code snippet handle this case.

I wondering whether a node can accept any group request and route to a correct node, an external router (lib) should take the responsibility.

huachaohuang:
The design is too simple to be merged so far. Let's leave it as a draft for now.

PR #689, opened 2022-05-24
Set up cargo workspace

huachaohuang:
Try to make it simple for now:

- `src/bin`: contains the `engula` binary
- `src/server`: contains the server side logic
- `src/engine`: contains the local storage engine


w41ter:
@huachaohuang Is local storage engine similar to rocksdb, or is it node or store? And where is the kv client lied in?

PR #693, opened 2022-05-24
api: add public API definition

### LINKED ISSUE #694: Should we add a crate to manage client related code?
The current crates designs does not reflect the works done by client. In fact, the code logic of client is very complex, including location cache, routing, error retries, txn ... At the same time, the client code itself will be depended on by other developers.

Therefore I think it is necessary to manage the client as a seperate crate. What is your opinion? @huachaohuang @zojw @tisonkun

huachaohuang:
Closes #694

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,2 @@
 [workspace]
-members = ["src/api", "src/bin", "src/engine", "src/server"]
+members = ["src/api", "src/bin", "src/client", "src/engine", "src/server"]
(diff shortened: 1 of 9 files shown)


w41ter (dismissed):

Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = ["src/bin", "src/engine", "src/server"]
+members = ["src/api", "src/bin", "src/engine", "src/server"]
```
tisonkun, line 2: ... also include `src/client`?

huachaohuang: Oh, yes.

PR #698, opened 2022-05-25
api: add more server api

huachaohuang:


src/api/engula/server/v1/root.proto
```diff
@@ -16,10 +16,12 @@ syntax = "proto3";
 
 package engula.server.v1;
 
+import "engula/v1/engula.proto";
 import "engula/v1/metadata.proto";
 import "engula/server/v1/metadata.proto";
 
 service Root {
+  rpc Admin(engula.v1.AdminRequest) returns (engula.v1.AdminResponse) {}
```
tisonkun, line 24: It seems that this method is added to `service Engula` already. What's the different between a `Engula` service and a `Root` service?

As physical concepts, it's reasonable to me that `service Root` and `service Node` implements these methods. However, who will implement `service Engula`? Whether it's implemented by every node or root node only, this method seems redundant.

The original case cause this question should be, if I implement a client, which destination I should send the `AdminRequest`?  A `Root` or an `Engula` service? What's an `Engula` service?

cc @zojw @w41ter

huachaohuang: There are two layers here, the client layer and the router layer. A client only talks to the Engula service and has no information about routing. A router receives requests from clients and routes them to individual nodes. A router talks to individual nodes through the Node and Root services.

tisonkun: What role will implement the so-called "Engula service"? Every Node?

huachaohuang: Yes, if we allow clients to connect to every node.

PR #701, opened 2022-05-27
server: bootstrap cluster

w41ter:


src/server/src/bootstrap.rs
```diff
@@ -35,18 +64,90 @@ async fn bootstrap_services(addr: &str, server: Server) -> Result<()> {
...
+    const FIRST_REPLICA_ID: u64 = 0;
+
+    // TODO(walter) write initial cluster data.
+    // - meta shards
+    let group = GroupDesc {
+        id: ROOT_GROUP_ID,
+        shards: vec![],
+        replicas: vec![],
+    };
+    node.create_replica(FIRST_REPLICA_ID, group).await?;
```
w41ter, line 140: @zojw We can initialize the meta range information after creating the replica here.

w41ter:
The next step is implement `Join` operation.

src/server/proto/v1/raft.proto
```diff
@@ -14,7 +14,7 @@
 
 syntax = "proto3";
 
-package engula.server.v1;
+package serverpb.v1;
```
huachaohuang, line 17: Maybe `engula.internal.v1`

huachaohuang (approved):

PR #702, opened 2022-05-30
server: execute GroupRequest

w41ter:


src/api/engula/v1/engula.proto
```diff
@@ -156,7 +156,7 @@ message CollectionResponseUnion {
 
 message GetRequest { bytes key = 1; }
 
-message GetResponse { bytes value = 1; }
+message GetResponse { optional bytes value = 1; }
```
huachaohuang, line 159: I think in "proto3", every field is optional by default.

src/api/engula/server/v1/node.proto
```diff
@@ -60,6 +65,14 @@ message GetRootRequest {}
 
 message GetRootResponse { repeated string addrs = 1; }
 
-message CreateReplicaRequest { GroupDesc group = 1; }
+message CreateReplicaRequest {
+  uint64 replica_id = 1;
```
huachaohuang, line 69: Can the node know which replica is being created from the node id in `ReplicaDesc`?

src/server/src/service/node.rs
```diff
@@ -38,6 +63,39 @@ impl node_server::Node for Server {
...
+            Some(replica) => replica,
+            None => {
+                return Err(Error::StaledRequest(request.group_id));
+            }
+        };
+        replica.execute(&request).await
+    }
+}
+
+fn error_2_response(err: Error) -> GroupResponse {
```
huachaohuang, line 90:
```suggestion
fn error_to_response(err: Error) -> GroupResponse {
```

src/api/engula/v1/engula.proto
w41ter: The basic types isn't optional by default. Here is the generated codes for rust:

```rust
#[derive(Clone, PartialEq, ::prost::Message)]
pub struct GetRequest {                                                                                                                                                               	 
    #[prost(bytes="vec", tag="1")]                                                                                                                                                 
    pub key: ::prost::alloc::vec::Vec<u8>,                                                                                                                                        }                                                                                                                                                                                 #[derive(Clone, PartialEq, ::prost::Message)]
pub struct GetResponse {                                                                                                                                                           
    #[prost(bytes="vec", optional, tag="1")]                                                                                                                                       
    pub value: ::core::option::Option<::prost::alloc::vec::Vec<u8>>,                                                                                                              }
```

Without `optional`, the type of `GetRequest::key` is `Vec<u8>, and `GetResponse::value` is `Option<Vec<u8>>`.

src/api/engula/server/v1/node.proto
w41ter: Some scenes can be found, but a few scenes can't be found. Because `replicas` represents a member of a raft group, and the newly created `replica` may not be in `replicas` (such as adding a new replica).

PR #709, opened 2022-05-31
api: add Join & CreateShard method

w41ter:


huachaohuang (approved):

src/api/engula/server/v1/root.proto
```diff
@@ -23,6 +23,10 @@ import "engula/server/v1/metadata.proto";
 service Root {
   rpc Admin(engula.v1.AdminRequest) returns (engula.v1.AdminResponse) {}
   rpc Watch(WatchRequest) returns (stream WatchResponse) {}
+
+  /// Join a bootstraped cluster. If the target node is itself not part of a
+  /// bootstraped cluster, an appropriate error is returned.
```
tisonkun, line 28:
```suggestion
  /// Join a bootstrapped cluster. If the target node is itself not part of a
  /// bootstrapped cluster, an appropriate error is returned.
```

@w41ter typo. You may push a commit to fix it :)

src/api/engula/server/v1/node.proto
```diff
@@ -50,6 +50,9 @@ message GroupRequestUnion {
     engula.v1.GetRequest get = 1;
     engula.v1.PutRequest put = 2;
     engula.v1.DeleteRequest delete = 3;
+
+    /// Add a new shard to an existing group.
+    CreateShardRequest create_shard = 4;
```
tisonkun, line 55: Hi @w41ter,

It seems a bit weird to me that CreateShard is at the same level with Get/Put/Delete. The flatten layout of such request is:

```rust
group_id: u64,
shard_id: u64,
request:
  create_shard:
    shard:
      id: u64,
      parent_id: u64,
      partition: ...
```

You may notice that shard id occurs multiple times and perhaps this request deserve a dedicated interface?

tisonkun: I'd prefer separate data and meta requests instead of putting all request into one super type and there are several _rules_ about which fields are relevant in specific situation.

Issue #708, opened 2022-05-31
setup root group metadata


zojw:

w41ter:
Close it since #718 has been merged.

PR #718, opened 2022-06-01
root: add root schema bootstrap

zojw:
ref #708, Bootstrap `_system_` database with 5 bootstrapping  collections:

- collection
- meta
- database
- node
- group

Then export some `Admin()` implements based on those collections.

### DIFF
--- a/src/server/src/bootstrap.rs
+++ b/src/server/src/bootstrap.rs
@@ -16,6 +16,7 @@ use std::{
     path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
+    vec,
 };
 
 use engula_api::server::v1::{
@@ -26,15 +27,23 @@ use tracing::{debug, info, warn};
 
 use crate::{
     node::{state_engine::StateEngine, Node},
+    root::Root,
     runtime::Executor,
     serverpb::v1::{raft_server::RaftServer, NodeIdent, ReplicaState},
     Error, Result, Server,
 };
 
 // TODO(walter) root cluster and first replica id.
-const ROOT_GROUP_ID: u64 = 0;
-const FIRST_REPLICA_ID: u64 = 0;
-const FIRST_NODE_ID: u64 = 0;
+pub const ROOT_GROUP_ID: u64 = 0;
+pub const NA_SHARD_ID: u64 = 0;
+pub const ROOT_SHARD_ID: u64 = 1;
+pub const FIRST_REPLICA_ID: u64 = 0;
+pub const FIRST_NODE_ID: u64 = 0;
+
+lazy_static::lazy_static! {
+    pub static ref MIN_KEY: Vec<u8> = vec![];
+    pub static ref MAX_KEY: Vec<u8> = vec![0xff, 0xff];
+}
 
 /// The main entrance of engula server.
 #[allow(unused)]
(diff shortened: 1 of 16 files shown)


src/server/src/root/schema.rs
```diff
@@ -0,0 +1,413 @@
...
+                    end: MAX_KEY.to_owned(),
+                })),
+            }],
+        })
+        .await?;
+
+        Ok(cluster_id)
+    }
+
+    pub async fn init_root_group_shard(&self) -> Result<()> {
```
w41ter, line 190: It is more appropriate to initialize the root group shard in `write_initial_cluster_data`, and directly add the shard to `GroupDesc` before calling `create_replica`.

Because the sign that the cluster has completed initialization is that the `NodeIdent` has been successfully persisted. The advantage of this is to ensure the atomicity of the initialization process.

If you call `store.create_shard` to create a `Shard` after the replica starts to serve, you must ensure that the user does not set the `--init` option after restarting, otherwise an error will be reported.

src/server/src/root/schema.rs
```diff
@@ -0,0 +1,413 @@
...
+            name: SYSTEM_GROUP_COLLECTION.to_owned(),
+            parent_id: SYSTEM_DATABASE_ID,
+        };
+        self.put_collection(group_collection.to_owned()).await?;
+        Ok(group_collection.id + 1)
+    }
+
+    async fn init_meta_collection(&self, next_collection_id: u64) -> Result<Uuid> {
+        let cluster_id = uuid::Uuid::new_v4();
+        self.put_meta(META_CLUSTER_ID_KEY.into(), cluster_id.as_bytes().to_vec())
```
w41ter, line 243: The `save_node_ident` function in bootstrap.rs also needs to fill in the cluster id。

src/server/src/root/schema.rs
```diff
@@ -0,0 +1,413 @@
...
+// bootstrap schema.
+impl Schema {
+    pub async fn try_bootstrap(&mut self, addr: &str) -> Result<Uuid> {
+        if let Some(cluster_id) = self.cluster_id().await? {
+            return Ok(cluster_id);
+        }
+
+        let next_collection_id = self.init_system_collections().await?;
+
+        let cluster_id = self.init_meta_collection(next_collection_id).await?;
```
w41ter, line 156: Like `init_root_group_shard`, initialization of other meta-information should also guarantee crash-recovery consistency.

For example, after `META_CLUSTER_ID_KEY` is persisted and restarted, subsequent ID keys such as `META_DATABASE_ID_KEY` have no chance to be created.

w41ter (commented):
Add some comments, the rest LGTM.

src/server/src/root/schema.rs
zojw: good catch 😄 , maybe we can use batch_put to ensure it?

I'd try to change them to use batch_put, implementation of batch_put should be in another pr https://github.com/engula/engula/pull/718/commits/e6de303436ac05b29887a37dde075a1ffe05ebe5#diff-07bcc1c599f14fc15c068c5d32dfd36f609b46d2072da4306004b0dd16a2258fR312

w41ter (approved):

src/server/src/bootstrap.rs
```diff
@@ -114,31 +128,29 @@ async fn bootstrap_or_join_cluster(
...
             "both cluster and node are initialized, node id {}",
             node_ident.node_id
         );
-        return Ok(());
+        return Ok(node_ident.cluster_id);
     }
 
-    if init {
-        bootstrap_cluster(node, addr).await?;
+    Ok(if init {
```
huachaohuang, line 141: I prefer the original style here.

huachaohuang (approved):

PR #720, opened 2022-06-06
server: poll root replica and root leader

w41ter:
According to the opinions communicated with @zojw , the root bootstrap has been modified.
1. Poll if the node has a root replica
2. Poll if the root replica is the leader

### DIFF
--- a/src/server/src/root/mod.rs
+++ b/src/server/src/root/mod.rs
@@ -16,68 +16,140 @@ mod job;
 mod schema;
 mod store;
 
-use std::sync::Arc;
+use std::{
+    sync::{Arc, Mutex},
+    task::Poll,
+    time::Duration,
+};
 
 use self::{schema::Schema, store::RootStore};
 use crate::{
-    node::Node,
+    node::{Node, Replica, ReplicaRouteTable},
     runtime::{Executor, TaskPriority},
     Error, Result,
 };
 
+#[derive(Clone)]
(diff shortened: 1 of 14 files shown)


zojw (approved):

tisonkun (approved):

src/server/src/root/mod.rs
```diff
@@ -16,68 +16,140 @@ mod job;
...
+        let mut bootstrapped = false;
+        loop {
+            let root_replica = self.fetch_root_replica(&replica_table).await;
 
-        self.refresh_root_owner(node);
-        self.executor
-            .spawn(None, TaskPriority::Middle, async move {}); // TODO(zojw): refresh owner, heartbeat node, rebalance
-        Ok(())
+            // Wait the current root replica becomes a leader.
+            if root_replica.on_leader().await.is_ok() {
```
zojw, line 88: Just mark a possible problem...

Now we are in a root follower node, it `fetch_root_replica` succeed with replica: "r1b" in L85, then wait "r1b" in L88, but "r1b" is removed and add new "r1e" in this node for some reason, and "r1e" even become leader,  it seems L88 has no chance to return from await...

maybe we can test this later when have time - -

w41ter: Cancel async task when drop is not supported yet, I will add it later. At that time, once the replica is removed, it will wake up immediately.

PR #731, opened 2022-06-08
node: support get_root

### LINKED ISSUE #730: implement `Node::GetRoot`

zojw:
closes #730

Support `get_root` on node service

1. First bootstrap  node, save itself as root node in its state_engine
2. Other join bootstrap node, save join response's  root nodes in state_engine
3. Root leader broadcast root nodes(and first element is current leader) to all known recent nodes periodically
4. Node handles heartbeat and update root info in memory + state_engine
6. Client should call get_node and try first element(high probability as current root leader)
7. Root service method should return NotRoot error with its known root

The node power-off long time and all its known root nodes have dead maybe not work until recieve root hearbeat

### DIFF
--- a/src/server/src/bootstrap.rs
+++ b/src/server/src/bootstrap.rs
@@ -171,12 +171,12 @@ async fn bootstrap_or_join_cluster(
     Ok(if init {
         bootstrap_cluster(node, addr).await?
     } else {
-        try_join_cluster(state_engine, addr, join_list).await?
+        try_join_cluster(node, addr, join_list).await?
     })
 }
 
 async fn try_join_cluster(
-    state_engine: &StateEngine,
+    node: &Node,
     local_addr: &str,
     join_list: Vec<String>,
 ) -> Result<NodeIdent> {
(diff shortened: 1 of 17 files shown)


src/server/src/bootstrap.rs
```diff
@@ -193,7 +194,7 @@ async fn try_join_cluster(
         ));
     }
 
-    let prev_cluster_id = if let Some(ident) = state_engine.read_ident().await? {
+    let prev_cluster_id = if let Some(ident) = node.state_engine().read_ident().await? {
```
w41ter, line 197: If a node needs to join to the cluster, it must have no node ident, otherwise it will be returned directly in bootstrap_or_join_cluster.

w41ter (approved):

huachaohuang (approved):

src/server/src/bootstrap.rs
huachaohuang: Should we fix it in this PR or later?

w41ter: I will submit a new PR to fix it.

PR #732, opened 2022-06-08
Bump rust toolchain to `nightly-2022-06-08`

leiysky:


tisonkun:
It seems new toolchain cause clippy complains :/

leiysky:
https://github.com/tokio-rs/prost/issues/661

The issue got closed as soon as I was going to vote for it. So let's just disable the clippy for all the generated code.

tisonkun (approved):
LGTM. @huachaohuang I'm unsure your preference on toolchain bump. My opinion is that as long as we select a concrete version (instead of "nightly" which will auto update) and all tests pass, such version bump does no harm and potentially ease the burden of version upgrading when we really meet some features only available in a new version.

+1 to merge from my side.

huachaohuang (approved):
I'm OK with this if it works fine.

PR #739, opened 2022-06-10
api,server: remove cluster_id in JoinNodeRequest

w41ter:
This filed was introduced in #718, it aims to prevent nodes that do not belong to the target cluster from being added to the cluster.

In the previous PRs(https://github.com/engula/engula/pull/731#discussion_r892999362, https://github.com/engula/engula/pull/737#discussion_r893666728), I found that this code does not actually execute, because each Node will only successfully call `join` once, and save the cluster id. Once successful, no subsequent `join` request will be issued.

This field was added to `JoinNodeRequest` by mistake, so I removed it. For its purpose, there should be a separate RPC (or adding a `cluster_id` field on the key RPC requests) to confirm that the cluster id matches.

### DIFF
--- a/src/api/engula/server/v1/root.proto
+++ b/src/api/engula/server/v1/root.proto
@@ -62,7 +62,6 @@ message WatchResponse {
 
 message JoinNodeRequest {
   string addr = 1;
-  optional bytes cluster_id = 2;
 }
 
 message JoinNodeResponse {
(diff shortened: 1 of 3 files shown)


w41ter:
> Could you write a brief description about what's the original purpose introduce this field and why we'll remove it now?

Updated.

huachaohuang (approved):

PR #754, opened 2022-06-15
root: support watch root change

### LINKED ISSUE #748: implement `Root::Watch`

zojw:
closes #748

### DIFF
--- a/src/server/src/root/mod.rs
+++ b/src/server/src/root/mod.rs
@@ -15,6 +15,7 @@
 mod job;
 mod schema;
 mod store;
+mod watch;
 
 use std::{
     sync::{Arc, Mutex},
@@ -22,8 +23,11 @@ use std::{
     time::Duration,
 };
 
-pub use self::schema::Schema;
 use self::store::RootStore;
+pub use self::{
+    schema::Schema,
+    watch::{WatchHub, Watcher, WriteGuard},
+};
 use crate::{
     node::{Node, Replica, ReplicaRouteTable},
     runtime::{Executor, TaskPriority},
--- a/src/server/src/root/mod.rs
+++ b/src/server/src/root/mod.rs
@@ -18,12 +18,22 @@ mod store;
 mod watch;
 
 use std::{
+    collections::HashMap,
     sync::{Arc, Mutex},
     task::Poll,
     time::Duration,
 };
 
-use self::store::RootStore;
+use engula_api::{
+    server::v1::{
+        report_request::GroupUpdates,
+        watch_response::{delete_event, update_event, DeleteEvent, UpdateEvent},
+        NodeDesc,
+    },
+    v1::{CollectionDesc, DatabaseDesc},
+};
+
+use self::{schema::ReplicaNodes, store::RootStore};
 pub use self::{
     schema::Schema,
     watch::{WatchHub, Watcher, WriteGuard},
@@ -45,6 +55,7 @@ struct RootShared {
     node_ident: NodeIdent,
     local_addr: String,
     core: Mutex<Option<RootCore>>,
+    watcher_hub: Arc<WatchHub>,
 }
 
 struct RootCore {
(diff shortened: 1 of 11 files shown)


src/server/src/root/mod.rs
```diff
@@ -158,8 +180,259 @@ impl Root {
...
+    use crate::{
+        bootstrap::bootstrap_cluster,
+        node::{Node, StateEngine},
+        root::Root,
+        runtime::{Executor, ExecutorOwner},
+        serverpb::v1::NodeIdent,
+    };
+
+    fn create_root(executor: Executor, node_ident: &NodeIdent) -> Root {
+        Root::new(executor, node_ident, "0.0.0.0:8888".into())
```
w41ter, line 351: A more appropriate way is to use an integration test to setup a listener, then use the client to connect and test. 

And a function `next_avail_port` in `src/server/test/helper/socket.rs` that can be used to find an available port to listen on. This avoids test failures due to port conflicts.

src/server/src/root/watch.rs
```diff
@@ -0,0 +1,180 @@
...
+
+pub struct WatcherInitializer<'a> {
+    _guard: RwLockWriteGuard<'a, WatchHubInner>,
+    watcher_inner: Arc<Mutex<WatcherInner>>,
+}
+
+impl<'a> WatcherInitializer<'a> {
+    pub fn set_init_resp(&mut self, updates: Vec<UpdateEvent>, deletes: Vec<DeleteEvent>) {
+        let mut inner = self.watcher_inner.lock().unwrap();
+        inner.init_resp = Some(WatchResponse { updates, deletes })
```
w41ter, line 50: Is it possible to append directly to `updates` and `deletes` ? In this way, `poll_next` does not need to check `init_resp` alone.

src/server/src/root/watch.rs
```diff
@@ -0,0 +1,180 @@
...
+    inner: Arc<std::sync::Mutex<WatcherInner>>,
+}
+
+#[derive(Default)]
+struct WatcherInner {
+    waker: Option<Waker>,
+    updates: Vec<UpdateEvent>,
+    deletes: Vec<DeleteEvent>,
+    err: Option<Error>,
+    dropped: bool,
```
w41ter, line 123: This field is unnecessary, because after the `Watcher` is released, the only thing that can be accessed is the `WatcherInitializer`, which will also be released later. After this, `poll_next` will not be called.

After `Poll::Ready(None)` returns, the stream will be closed, so you can use `err` as the basis for whether to close the stream.

w41ter (dismissed):
There are some comments here, reset LGTM, but let's approve first and move the development forward.

src/server/src/root/watch.rs
zojw: watcher(inner) may still be reachable after `Watcher` is released, because it still in `WatchHub.inner.watchers` hashmap.

A more appropriate way should remove itself from watchers hashmap when watcher `Drop`， but it seems we cannot call async lock/unlock inside `Drop`, so it choose set a drop mark field, and cleanup in other place(watchhub cleanup...emm it's no caller now- -)

src/server/src/root/mod.rs
zojw: good idea, I will try it later, root still need more test

src/server/src/root/watch.rs
w41ter: This should be a mistake, considering that only the stream where the error occurred will be closed, then the watcher should be removed from the `WatcherHub` when `notify` with error.

In addition, if you want to notify `WatchHub` to delete the `Watcher` after the stream is closed by the client, it should be designed as `Sender` and `Receiver` mode, they share a common structure, and `WatchHub` keep `Sender`, `Stream` keep `Receiver`, in this case drop is useful. Otherwise drop will not be called as long as `WatcherHub` still holds `Watcher`.

w41ter (approved):

src/server/src/root/watch.rs
w41ter: I got it wrong, `Stream` and `WatchHub` hold different `Watcher`s, so `Drop` can be called after `Stream` is closed.

PR #774, opened 2022-06-20
root: save system db as "one shard per collection" & support list system collection

### LINKED ISSUE #773: support prefix list in a shard

zojw:
as pre-step of #766, it makes collections in `__system__` as "one init-shard per collection" and support list values in the collection(it also assumes that `__system__` database's shard doesn't need auto split)

closes #773

### DIFF
--- a/src/server/src/node/engine/group.rs
+++ b/src/server/src/node/engine/group.rs
@@ -247,17 +247,21 @@ impl GroupEngine {
         Ok(())
     }
 
-    pub fn iter(&self) -> Result<GroupEngineIterator> {
-        use rocksdb::{IteratorMode, ReadOptions};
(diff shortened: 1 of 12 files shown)


src/server/src/node/engine/group.rs
```diff
@@ -247,17 +247,21 @@ impl GroupEngine {
         Ok(())
     }
 
-    pub fn iter(&self) -> Result<GroupEngineIterator> {
-        use rocksdb::{IteratorMode, ReadOptions};
+    pub fn iter(&self, from: Option<Vec<u8>>) -> Result<GroupEngineIterator> {
```
huachaohuang, line 250: I prefer adding another interface like `iter_from` instead of forcing users to pass a `None` here.

huachaohuang (approved):
Rest LGTM

huachaohuang (approved):

PR #781, opened 2022-06-22
feat(client): add engula_client

### LINKED ISSUE #765: Add router that holds state to lookup nodes
#752 added a basic structure of router. We should finish the `todo`s and build a state for looking up nodes based on the database, the collection, and the key to manipulate.

tisonkun:
This closes #765.


tisonkun:
You can run for debugging with:

```shell
cargo run start --db <db-path> --init # start server
```

```shell
cargo run --example kv
```

src/client/src/app_client.rs
```diff
@@ -0,0 +1,316 @@
...
+                        RequestBatchBuilder::new(node_id)
+                            .get(group.id, epoch, shard.id, key)
+                            .build(),
+                    )
+                    .await?
+            }
+        };
+
+        for r in resp {
+            if let Some(err) = r.error {
```
w41ter, line 303: The following errors need to be handled:
- `EpochNotMatch`
- `GroupNotFound`
- `NotRootLeader`
- `NotLeader`

The previous `put` also needs to handle errors in the same way.

w41ter (commented):
Rest LGTM.

src/client/src/app_client.rs
tisonkun: There is another error handling logic I don't have a clear idea yet: If we don't find shard as currently it may be a bug, shall we retry also since it's also possible that a group event is not synced yet?

Or may you comment in details how to handle errors you mentioned? For example, I don't know what the client can do if it get `GroupNotFound`.

w41ter (approved):

zojw (approved):
LGTM, it also can pass test with #783

tisonkun:
Thanks for your review! Merging...

I'll verify #783 now.

PR #783, opened 2022-06-22
server: make server testable with client

zojw:
tiny fix to pass #781 client example test

### DIFF
--- a/src/server/src/node/job/report_state.rs
+++ b/src/server/src/node/job/report_state.rs
@@ -114,6 +114,7 @@ async fn report_state_updates(roots: &mut Vec<String>, request: ReportRequest) {
 /// If one day you find that reporting has become a bottleneck, you can consider optimizing this
 /// code.
 async fn issue_report_request(addr: String, request: &ReportRequest) -> Result<()> {
+    let addr = format!("http://{}", addr);
     let mut client = RootClient::connect(addr).await?;
     client.report(request.clone()).await?;
     Ok(())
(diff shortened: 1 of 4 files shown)


tisonkun:
@zojw can you rebase this patch onto latest main branch?

zojw:
> @zojw can you rebase this patch onto latest main branch?

done~

tisonkun (approved):
Verified that we can run a minimal KV put/get example. +1 to merge.

May I ask what the change you made in this patch to fix the issue? It's not quite straightforward from the diff itself.

w41ter (approved):

src/server/src/node/job/report_state.rs
```diff
@@ -114,6 +114,7 @@ async fn report_state_updates(roots: &mut Vec<String>, request: ReportRequest) {
 /// If one day you find that reporting has become a bottleneck, you can consider optimizing this
 /// code.
 async fn issue_report_request(addr: String, request: &ReportRequest) -> Result<()> {
+    let addr = format!("http://{}", addr);
```
tisonkun, line 117: Weird. I think that RootClient and effectively NodeClient should prepend `"http://"`.

```rust
// RootClient
pub async fn connect(addr: String) -> Result<Self, crate::Error> {
    let node_client = NodeClient::connect(addr).await?;
    // ...
}

// NodeClient
pub async fn connect(addr: String) -> Result<Self, tonic::transport::Error> {
    let addr = format!("http://{}", addr);
    // ...
}
```

zojw: yes, `engula_client::RootClient` can handle it rightly, but these methods use protobuf's RootClient and self-maintain root address list and miss it

tisonkun: Well. This `RootClient` is from protobuf generated :P

PR #797, opened 2022-06-27
root: handle heartbeat

### LINKED ISSUE #766: send heartbeat and collect state & stats

zojw:
closes #766

### DIFF
--- a/src/server/src/node/mod.rs
+++ b/src/server/src/node/mod.rs
@@ -478,9 +478,7 @@ impl Node {
                 }
 
                 let state = replica.replica_state();
-                if state.role == RaftRole::Leader.into() {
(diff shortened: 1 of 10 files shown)


w41ter (dismissed):
LGTM

src/server/src/node/mod.rs
```diff
@@ -477,9 +477,7 @@ impl Node {
                 }
 
                 let state = replica.replica_state();
-                if state.role == RaftRole::Leader.into() {
```
w41ter, line 480: In the previous design, only the leader would report group descriptors. What is the purpose of this change?

w41ter (approved):

PR #798, opened 2022-06-28
server: iterate within shard

### LINKED ISSUE #796: iterate within shard
The current iterator implementation only verify the collection id: if there are multiple shards in a group, the iterator will return all the data of these shards. 

After the shard migration is complete, iterator will be used to scan and delete all keys of the source shard, which requires the iterator only returning datas in the source shard. Therefore, the scope of iterator can only be a shard.

w41ter:
This PR closes #796.

### DIFF
--- a/src/server/src/node/engine/group.rs
+++ b/src/server/src/node/engine/group.rs
@@ -637,7 +637,7 @@ impl SnapshotRange {
         match self {
             SnapshotRange::Target { target_key } if target_key == key => true,
             SnapshotRange::Prefix { prefix } if key.starts_with(prefix) => true,
-            SnapshotRange::Range { start, end } if shard::in_range(&start, &end, key) => true,
+            SnapshotRange::Range { start, end } if shard::in_range(start, end, key) => true,
             _ => false,
         }
     }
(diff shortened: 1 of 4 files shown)


src/server/src/node/engine/group.rs
```diff
@@ -366,74 +336,80 @@ impl GroupEngine {
...
+        self.shard_descs
             .read()
             .expect("read lock")
             .get(&shard_id)
             .cloned()
+            .ok_or_else(|| Error::InvalidArgument(format!("no such {} shard exists", shard_id)))
+    }
+
+    #[inline]
+    fn cf_handle(&self) -> Arc<rocksdb::BoundColumnFamily> {
```
huachaohuang, line 409: Maybe you can just store it as a field.

huachaohuang (approved):

src/server/src/node/engine/group.rs
w41ter: > Maybe you can just store it as a field.

Not easy, the `BoundColumnFamily` requires a lifetime parameter which equals to lifetime of `rocksdb::DB`. To store it as a field, a self-referenced structure must be build.

PR #801, opened 2022-06-28
server: migration state transitions

### LINKED ISSUE #800: migration state transitions

w41ter:
This PR closes #800.


w41ter:
Because the column family ID is not considered in the design of `WriteBatchRep`, there will be an error when `write` the leader's WriteBatch on the follower's rocksdb, so the multi-replicas shard migration test is temporarily disabled.

huachaohuang (approved):

PR #802, opened 2022-06-29
server: fix initialize raft group

w41ter:
The underlying `raft-rs` crate isn't allow empty configs enter joint state, so have to add replicas one by one.


huachaohuang (approved):

huachaohuang:
@w41ter CI failed

PR #806, opened 2022-06-29
client: resume watching event stream

### LINKED ISSUE #804: resume watching events stream

w41ter:
This PR closes #804.

### DIFF
--- a/src/client/src/router.rs
+++ b/src/client/src/router.rs
@@ -15,6 +15,7 @@
 use std::{
     collections::HashMap,
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
 use engula_api::{server::v1::*, v1::*};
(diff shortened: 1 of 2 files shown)


src/client/src/router.rs
```diff
@@ -109,11 +107,50 @@ impl Router {
...
+        match RootClient::connect(addr.clone()).await {
+            Ok(c) => break c,
+            Err(e) => {
+                warn!(err = ?e, addr=?addr, "connect root server");
+            }
+        };
+
+        tokio::time::sleep(Duration::from_millis(interval)).await;
+        interval = std::cmp::min(interval * 2, 1000);
+    };
```
tisonkun, line 122: I'd suggest you just create `root_client` in the loop below. Retry here is clumsy since we lookup root by `get_root` inside `RootClient::connect` - retry doesn't effectively change conditions.

w41ter: In the node server, the cluster may not have completed bootstrap when the router is created. So the retry here is used to handle the failure of the underlying `NodeClient::connect()`.

tisonkun: OK.

tisonkun (approved):

PR #845, opened 2022-07-13
root: add some root diagnosis log

zojw:
ref #834

### DIFF
--- a/src/server/src/error.rs
+++ b/src/server/src/error.rs
@@ -25,6 +25,15 @@ pub enum Error {
     #[error("database {0} not found")]
     DatabaseNotFound(String),
 
+    #[error("database {0} exist")]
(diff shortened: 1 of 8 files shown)


huachaohuang:
Hmm, we got invalid pointer in Rust (https://github.com/engula/engula/runs/7322786704?check_suite_focus=true#step:8:85) 🤔

w41ter:
> Hmm, we got invalid pointer in Rust ([engula/engula/runs/7322786704?check_suite_focus=true#step:8:85](https://github.com/engula/engula/runs/7322786704?check_suite_focus=true#step:8:85)) 🤔

This is caused by thread exiting, because the TestContext haven't graceful shutdown right now.

w41ter (approved):

PR #867, opened 2022-07-19
root: refine root allocator 

### LINKED ISSUE #861: root allocator continue try transfer leader & replica to shutdown node
1. setup 4 node cluster (n1, n2, n3, n4)
2. keep n1 as the root leader and wait for cluster balanced for replica & leader cnt
3. shutdown n4
4. group cure logic will repair lost replica in n4

but the allocator still tries to transfer leader / replica to unreachable node n4

allocator need take care node_livess when compute balance action

zojw:
1. avoid moving or advising replica to the node that already contains replica(even if the one that does not include in group_desc)

ref #856, this can relieve the recreating replica problem

3.  avoid useless drop root leader

turn off a normal node shouldn't observe drop leader log in root leader node.

4. add rough root liveness

closes #861, we can observe allocate stop trying to reallocate replica or transfer leader to offline-node after 1 min.

5. send heartbeat parallel

but slow reconcile still can cause inaccurate heartbeat interval, but it can be solve in next PR(add scheduler to async execute task)

### DIFF
--- a/src/server/src/root/job.rs
+++ b/src/server/src/root/job.rs
@@ -61,12 +61,19 @@ impl Root {
             });
         }
 
+        let mut hb_futs = Vec::new();
         for n in nodes {
             trace!(node = n.id, target = ?n.addr, "attempt send heartbeat");
-            match self
-                .try_send_heartbeat(&n.addr, &piggybacks, self.liveness.heartbeat_timeout)
-                .await
-            {
+            let fut = self.try_send_heartbeat(
+                n.addr.to_owned(),
+                &piggybacks,
+                self.liveness.heartbeat_timeout,
+            );
+            hb_futs.push((n.to_owned(), fut));
+        }
+
+        for (n, fut) in hb_futs {
+            match fut.await {
                 Ok(res) => {
                     self.liveness.renew(n.id);
                     for resp in res.piggybacks {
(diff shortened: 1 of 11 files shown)


src/server/src/root/job.rs
```diff
@@ -61,10 +61,21 @@ impl Root {
...
+            let fut = self.try_send_heartbeat(
+                n.addr.to_owned(),
+                &piggybacks,
+                self.liveness.heartbeat_timeout,
+            );
+            hb_futs.push((n.to_owned(), fut));
+        }
+
+        for (n, fut) in hb_futs {
+            match fut.await {
```
w41ter, line 76: It seems that here is not executed in parallel, because `poll` is called only when use `.await`, so it is still executed in sequential.

src/server/src/root/liveness.rs
```diff
@@ -0,0 +1,112 @@
...
+        const LIVENESS_FRAC: f64 = 0.5;
+
+        // ensure liveness_threshold >>> raft_election_timeout?
+        let liveness_threshold = Duration::from_millis(
+            cfg.raft.election_tick as u64
+                * cfg.raft.tick_interval_ms
+                * RAFT_ELECTION_TIMEOUT_MULTIPLIER,
+        );
+        let heartbeat_timeout =
+            Duration::from_millis((liveness_threshold.as_millis() as f64 * LIVENESS_FRAC) as u64);
```
w41ter, line 59: Actual raft election timeout is an randomly value in `[election_tick, 2 * election_tick]`.

w41ter (approved):

w41ter (approved):

PR #888, opened 2022-07-21
root: refactor reconcile logic as scheduler

### LINKED ISSUE #868: cleanup orphan replicas caused by failed create new group

zojw:
closes #868 

- restruct reconcile execution logic like node's scheduler
- reuse group_client to handle retriable-error


src/server/proto/v1/metadata.proto
```diff
@@ -146,6 +147,83 @@ message Migration {
 
   engula.server.v1.MigrationDesc migration_desc = 2;
 
-  /// The latest migrated key, used for fault tolerance, locate the cursor that has been replicated.
+  /// The latest migrated key, used for fault tolerance, locate the cursor that
+  /// has been replicated.
   bytes last_ingested_key = 3;
 }
+
+message ReconcileTask {
```
w41ter, line 155: How about move these definitions to `schedule.proto`?

w41ter (dismissed):

w41ter (approved):

w41ter:
It seems that the problem in CI is not caused by this PR, so let's merging it first and then fix problems in others PR.

PR #912, opened 2022-07-26
*: speed up ci building

w41ter:


tisonkun (approved):

tisonkun:
`cargo fmt` failed. But we don't touch rust files in this patch. How can the main branch be in such situation 🤣

w41ter:
My mistake, the ci is broken because I added some unstable tests, then I merge a request without validated the format... 🤣

PR #942, opened 2022-08-02
server: add process metrics

w41ter:
Refs #837


w41ter:
My mistake and I will closes this PR.

PR #997, opened 2022-08-16
*: upgrade deps

w41ter:
updated: tonic,prost,config-rs


w41ter:
Prost removed its own protoc in 0.8, which requires each development environment to have protoc installed. In order to avoid unnecessary trouble, I plan to only update config.rs

PR #1101, opened 2022-09-28
client: fix blocked on recreate pull stream

zojw:
fix blocked on recreating pull stream

also, add an additional leader check before the return stream in the node

### DIFF
--- a/src/server/src/node/mod.rs
+++ b/src/server/src/node/mod.rs
@@ -446,6 +446,7 @@ impl Node {
                 return Err(Error::GroupNotFound(request.group_id));
             }
         };
+        replica.check_migrating_request_early(request.shard_id)?;
         Ok(ShardChunkStream::new(
             request.shard_id,
             self.cfg.shard_chunk_size,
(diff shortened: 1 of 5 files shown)


src/server/src/node/mod.rs
```diff
@@ -446,6 +446,7 @@ impl Node {
                 return Err(Error::GroupNotFound(request.group_id));
             }
         };
+        replica.check_migrating_request_early(request.shard_id)?;
```
w41ter, line 449: Setting `check_migrating_request_early` to pub and accessing it directly may not be appropriate, since it doesn't get an acl guard and internally accesses fields that might be accessed by unsubmitted requests. Here `on_leader` should be enough.

w41ter: No big deal for now, later I want to enhance `ShardPrefixList` and replace `fetch_shard_chunk` with it, so it can be merged first and move forward.

w41ter (approved):

